PR #12, opened 2014-03-07
Changed print to use the result parameter from lldb (AppCode compatibility)

alistra:
This is needed for the outputs to be shown in AppCode, as with vanilla print it doesn't show anything


kastiglione:
https://youtrack.jetbrains.com/issue/OC-9356 has been marked as fixed.

PR #55, opened 2014-11-21
Feature: Accessibility Tree printing

KingOfBrian:
This is not quite ready, and full of terrible python, but the basics are there.   It's sort of comical how long it takes to run and the amount of commands sent to generate the output.   But I think it will be helpful with some tuning.  Any suggestions would be good to get.

It adds a pae command that prints out the accessibility tree starting from the current object.   If the current object is a UIView, it will also traverse subviews to find accessibility elements, as I *think* that's what voice over is doing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,8 @@
 
 [[Installation](#installation) &bull; [Commands](#commands) &bull; [Custom Commands](#custom-commands) &bull; [Development Workflow](#development-workflow) [Contributing](#contributing) &bull; [License](#license)]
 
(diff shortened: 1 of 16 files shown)


kastiglione:
@KingOfBrian Is this pull request superseded by #70? Or is there functionality here that's missing in #70?

kastiglione:
Closing since I understand this to be superseded by #70. @KingOfBrian if this is incorrect, feel free to re-open.

Thanks so much for your time, effort and contribution! :gem:

KingOfBrian:
Yea, no worries, glad we got something better in.  I may add some thing to print out more information when I have some ax work, but thanks for following up.

PR #67, opened 2014-12-24
Added 'slow' command.

bartoszj:
This command slows down animations. Works like 'Toggle Slow Animations' from iOS Simulator but this command works also on the device and user is able to select animation speed.

### DIFF
--- a/commands/FBDisplayCommands.py
+++ b/commands/FBDisplayCommands.py
@@ -24,6 +24,7 @@ def lldbcommands():
     FBUnmaskViewCommand(),
     FBShowViewCommand(),
     FBHideViewCommand(),
+    FBSlowAnimation()
   ]
 
 
@@ -145,3 +146,17 @@ class FBHideViewCommand(fb.FBCommand):
 
   def run(self, args, options):
     viewHelpers.setViewHidden(args[0], True)
+
+
+class FBSlowAnimation(fb.FBCommand):
+  def name(self):
+    return 'slow'
+
+  def description(self):
+    return 'Slows down animations. Works on the iOS Simulator and a device.'
+
+  def args(self):
+    return [ fb.FBCommandArgument(arg='speed', type='float', default='1', help='Animation speed.') ]
+
+  def run(self, args, option):
+    lldb.debugger.HandleCommand('expr [(CALayer *)[[[[UIApplication sharedApplication] windows] objectAtIndex:0] layer] setSpeed:(CGFloat)%s]' % (args[0]))
--- a/commands/FBDisplayCommands.py
+++ b/commands/FBDisplayCommands.py
@@ -24,7 +24,8 @@ def lldbcommands():
     FBUnmaskViewCommand(),
     FBShowViewCommand(),
     FBHideViewCommand(),
-    FBSlowAnimation()
+    FBSlowAnimationCommand(),
+    FBUnslowAnimationCommand()
   ]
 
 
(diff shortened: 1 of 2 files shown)


kastiglione:
Thank you @bartoszj :+1: 

bartoszj:
Done.

commands/FBDisplayCommands.py
```diff
@@ -145,3 +146,17 @@ def args(self):
 
   def run(self, args, options):
     viewHelpers.setViewHidden(args[0], True)
+
+
+class FBSlowAnimation(fb.FBCommand):
+  def name(self):
+    return 'slow'
```
kastiglione: What do you think about calling this `slowanim`?

commands/FBDisplayCommands.py
```diff
@@ -145,3 +146,17 @@ def args(self):
...
+
+  def description(self):
+    return 'Slows down animations. Works on the iOS Simulator and a device.'
+
+  def args(self):
+    return [ fb.FBCommandArgument(arg='speed', type='float', default='1', help='Animation speed.') ]
+
+  def run(self, args, option):
+    lldb.debugger.HandleCommand('expr [(CALayer *)[[[[UIApplication sharedApplication] windows] objectAtIndex:0] layer] setSpeed:(CGFloat)%s]' % (args[0]))
```
kastiglione: Can we run this for all windows instead of just the one at index 0?

kastiglione:
@bartoszj Thanks for submitting this! I have a couple comments above.

What do you think about adding two commands, one to turn this on, one to turn if off. All of our other commands work this way, so we should be consistent unless there's a good reason not to be. If we have two commands, they can both have a default speed, with the undo command having a speed of 1.

arigrant:
Isn't this the same as 'slowanim' and 'unslowanim' or have we not merged that, @kastiglione ?

bartoszj:
@kastiglione I agree with two commands. `slowanim` and `unslowanim` would be OK? Also I'll check slow animation on all windows.

kastiglione:
Great! :bow:

bartoszj:
Updated.

fblldbviewhelpers.py
```diff
@@ -87,3 +87,6 @@ def upwardsRecursiveDescription(view, maxDepth=0):
     currentPrefix += "   | "
 
   return builder
+
+def slowAnimation(speed=1):
+  lldb.debugger.HandleCommand('expr -- for (UIWindow *$w in (NSArray *)[[UIApplication sharedApplication] windows]) { [[$w layer] setSpeed:(CGFloat)%s]; }' % (speed))
```
kastiglione: This is awesome!

For what it's worth, another way that might have worked is to use `NSArray`'s custom implementation of `-setValue:forKey:`:

```objc
[[[UIApplication sharedApplication] windows] setValue:@(speed) forKey:@"speed"]
```

kastiglione: Where allowable, can we change the `XYClass *` casts to simply `id`?

commands/FBDisplayCommands.py
```diff
@@ -145,3 +147,28 @@ def args(self):
...
+    return 'slowanim'
+
+  def description(self):
+    return 'Slows down animations. Works on the iOS Simulator and a device.'
+
+  def args(self):
+    return [ fb.FBCommandArgument(arg='speed', type='float', help='Animation speed.') ]
+
+  def run(self, args, option):
+    viewHelpers.slowAnimation(args[0])
```
kastiglione: Can we provide a default? If I call `slowanim`, maybe set it to 0.25? Perhaps we can dig up the speed the simulator uses, but that's hardly worth spending much time on.

kastiglione:
Thanks @bartoszj! I left a couple comments, the main one is about giving `slowanim` a default speed value so that it could be typed without having to provide an explicit speed.

bartoszj:
I've found that iOS Simulator is using 0.1 as speed value and I've set it in the`slowanim` command.

kastiglione:
:boom: 

Looks good. I'll checkout this branch and give it a try and merge if all goes well. Thanks!

kastiglione:
Thanks for this contribution! And thanks for being patient and awesome :+1:

PR #68, opened 2014-12-26
Add basic Swift support (requires Xcode 6)

mblsha:
By default expressions are evaluated in current frame's language.

When stopped inside a Swift function, current language is Swift, and it's invalid to use Objective-C syntax: many Chisel functions break as a result.

Explicitly specify Objective-C++ as expression language when evaluating commands.

However one problem will still persist: it's still impossible to use Swift objects by name in Chisel functions. One needs to first convert object names to id pointers using evaluateExpressionValueInFrameLanguage(), but the lldb code to enable that was committed very recently and is still unavailable in Xcode 6.1.1.

### DIFF
--- a/fblldbbase.py
+++ b/fblldbbase.py
@@ -36,13 +36,26 @@ class FBCommand:
     pass
 
 
-def evaluateExpressionValue(expression, printErrors=True):
+def evaluateExpressionValueWithLanguage(expression, language, printErrors):
   # lldb.frame is supposed to contain the right frame, but it doesnt :/ so do the dance
   frame = lldb.debugger.GetSelectedTarget().GetProcess().GetSelectedThread().GetSelectedFrame()
-  value = frame.EvaluateExpression(expression)
+  expr_options = lldb.SBExpressionOptions()
+  expr_options.SetLanguage(language)  # requires lldb r210874 (2014-06-13) / Xcode 6
+  value = frame.EvaluateExpression(expression, expr_options)
   if printErrors and value.GetError() is not None and str(value.GetError()) != 'success':
     print value.GetError()
   return value
+
+def evaluateExpressionValueInFrameLanguage(expression, printErrors=True):
+  # lldb.frame is supposed to contain the right frame, but it doesnt :/ so do the dance
+  frame = lldb.debugger.GetSelectedTarget().GetProcess().GetSelectedThread().GetSelectedFrame()
+  language = frame.GetCompileUnit().GetLanguage()  # requires lldb r222189 (2014-11-17)
+  return evaluateExpressionValueWithLanguage(expression, language, printErrors)
+
+# evaluates expression in Objective-C++ context, so it will work even for
+# Swift projects
+def evaluateExpressionValue(expression, printErrors=True):
+  return evaluateExpressionValueWithLanguage(expression, lldb.eLanguageTypeObjC_plus_plus, printErrors)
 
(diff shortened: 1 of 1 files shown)


kastiglione:
Thank you @mblsha :octocat: 

mblsha:
Ok, done ;-)

fblldbbase.py
```diff
@@ -36,14 +36,27 @@ def run(self, arguments, option):
     pass
 
 
-def evaluateExpressionValue(expression, printErrors=True):
+def evaluateExpressionValueWithLanguage(expression, language, printErrors):
   # lldb.frame is supposed to contain the right frame, but it doesnt :/ so do the dance
   frame = lldb.debugger.GetSelectedTarget().GetProcess().GetSelectedThread().GetSelectedFrame()
-  value = frame.EvaluateExpression(expression)
+  expr_options = lldb.SBExpressionOptions()
+  expr_options.SetLanguage(language)  # requires lldb r210874 (2014-06-13) / Xcode 6
```
kastiglione: Hey @arigrant, what do you think about baseline support? Should Chisel support anything more than just the current version of Xcode?

fblldbbase.py
```diff
@@ -36,14 +36,27 @@ def run(self, arguments, option):
...
+  value = frame.EvaluateExpression(expression, expr_options)
   if printErrors and value.GetError() is not None and str(value.GetError()) != 'success':
     print value.GetError()
   return value
 
+def evaluateExpressionValueInFrameLanguage(expression, printErrors=True):
+  # lldb.frame is supposed to contain the right frame, but it doesnt :/ so do the dance
+  frame = lldb.debugger.GetSelectedTarget().GetProcess().GetSelectedThread().GetSelectedFrame()
+  language = frame.GetCompileUnit().GetLanguage()  # requires lldb r222189 (2014-11-17)
```
kastiglione: @mblsha Which version of Xcode includes `r222189`?

kastiglione:
@mblsha I've left a couple comments. Thanks for submitting this pull request!

The only thing that needs to be figured out Xcode/lldb version support.

fblldbbase.py
mblsha: Not sure, it's not yet in 6.1.1, so `evaluateExpressionValueInFrameLanguage()` is not actually called anywhere and is present for future reference. Hopefully LLDB in 6.2 beta has it, but I haven't verified it.

I'll try it once it becomes final.

mblsha:
`evaluateExpressionValueWithLanguage()` works fine in Xcode 6.0+. Once 6.2 becomes final maybe `evaluateExpressionValueInFrameLanguage()` will work as well.

fblldbbase.py
kastiglione: > `evaluateExpressionValueInFrameLanguage()` is not actually called anywhere

:+1:

moreindirection:
It looks like there are still a few problems in Swift. Any commands that are of the form "po EXPRESSION" need to be changed to "expression -O -l objc++ --- EXPRESSION"

Would really like to see these changes merged into master, though.

kastiglione:
:ship: :ship: :ship: Thanks @mblsha!

kastiglione:
@moreindirection Can I interest you in opening a pull request to fix the stuff you've seen broken with Swift? :cake:

moreindirection:
@kastiglione Sure, I just created pull request #73

PR #69, opened 2015-01-02
support x86_64h

samjohn:
Using OSX Yosemite 10.10.2  and Xcode Version 6.1.1 (6A2008a) my architecture comes up as x86_64h. This leads to the error: `Your architecture, {}, is truly fantastic. However, I don\'t currently support it.`

Adding x86_64h as an option to this file allows chisel to work properly.

### DIFF
--- a/fblldbobjcruntimehelpers.py
+++ b/fblldbobjcruntimehelpers.py
@@ -47,7 +47,7 @@ def functionPreambleExpressionForSelf():
   expressionForSelf = None
   if arch == 'i386':
     expressionForSelf = '*(id*)($esp+4)'
-  elif arch == 'x86_64':
+  elif arch == 'x86_64' or 'x86_64h':
     expressionForSelf = '(id)$rdi'
   elif arch == 'arm64':
     expressionForSelf = '(id)$x0'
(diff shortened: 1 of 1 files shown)


kastiglione:
Thanks @samjohn :star:

samjohn:
done

fblldbobjcruntimehelpers.py
```diff
@@ -47,7 +47,7 @@ def functionPreambleExpressionForSelf():
   expressionForSelf = None
   if arch == 'i386':
     expressionForSelf = '*(id*)($esp+4)'
-  elif arch == 'x86_64':
+  elif arch == 'x86_64' or 'x86_64h':
```
kastiglione: Instead of checking for variations at these call sites, would you mind doing a normalization inside `currentArch()`. In other words, make `currentArch()` return `'x86_64'` instead of `'x86_64h'`. Thoughts?

samjohn: yeah, I can do that. I wasn't sure if there were any significant differences between x86_64 and x86_64h that would need to be addressed elsewhere in the code. This fix works for bmessage but I haven't tried out all the functionality in the repo so it may still break in other places.

kastiglione:
:ship:

kastiglione:
Thanks again!

PR #70, opened 2015-01-07
Added print accessibility tree command. Find accessibility element command reworked

gkassabli:
Added print accessibility tree command that is based on Apple internal methods. It could print whole tree or subtree of specified root element.
Find accessibility element command was reworked to use internal API and make use of only relevant and visible elements

### DIFF
--- a/commands/FBAccessibilityCommands.py
+++ b/commands/FBAccessibilityCommands.py
@@ -49,14 +49,14 @@ class FBFindViewByAccessibilityLabelCommand(fb.FBCommand):
 
   def run(self, arguments, options):
     forceStartAccessibilityServer()
-    rootView = fb.evaluateExpression('(id)[[UIApplication sharedApplication] keyWindow]')
-    accessibilityGrepHierarchy(rootView, arguments[0])
+    rootView = fb.evaluateObjectExpression('[[UIApplication sharedApplication] keyWindow]')
+    accessibilityGrepHierarchy(rootView, arguments[0], False)
 
 def forceStartAccessibilityServer():
   #We try to start accessibility server only if we don't have needed method active
   if not fb.evaluateBooleanExpression('[UIView instancesRespondToSelector:@selector(_accessibilityElementsInContainer:)]'):
     #Starting accessibility server is different for simulator and device
-    if fb.evaluateBooleanExpression('((NSRange)[(NSString *)[(id)[(id)[UIDevice currentDevice] model] lowercaseString] rangeOfString:@"simulator"]).location != __LONG_MAX__'):
+    if fb.evaluateExpressionValue('(id)[[UIDevice currentDevice] model]').GetObjectDescription().lower().find('simulator') >= 0:
       lldb.debugger.HandleCommand('expr (void)[[UIApplication sharedApplication] accessibilityActivate]')
     else:
       lldb.debugger.HandleCommand('expr (void)[[[UIApplication sharedApplication] _accessibilityBundlePrincipalClass] _accessibilityStartServer]')
@@ -74,26 +74,26 @@ def printAccessibilityHierarchy(view, indent = 0):
   if int(a11yLabel.GetValue(), 16) == 0:
     print indentString + ('{} {}'.format(classDesc, view))
     #We call private method that gives back all visible accessibility children for view
-    accessibilityElements = fb.evaluateExpression('(id)[[[UIApplication sharedApplication] keyWindow] _accessibilityElementsInContainer:0 topLevel:%s includeKB:0]' % view)
-    accessibilityElementsCount = int(fb.evaluateExpression('(int)[%s count]' % accessibilityElements))
+    accessibilityElements = fb.evaluateObjectExpression('[[[UIApplication sharedApplication] keyWindow] _accessibilityElementsInContainer:0 topLevel:%s includeKB:0]' % view)
+    accessibilityElementsCount = fb.evaluateIntegerExpression('[%s count]' % accessibilityElements)
     for index in range(0, accessibilityElementsCount):
-      subview = fb.evaluateExpression('(id)[%s objectAtIndex:%i]' % (accessibilityElements, index))
+      subview = fb.evaluateObjectExpression('[%s objectAtIndex:%i]' % (accessibilityElements, index))
       printAccessibilityHierarchy(subview, indent + 1)
   else:
     print indentString + ('({} {}) {}'.format(classDesc, view, a11yLabel.GetObjectDescription()))
 
-def accessibilityGrepHierarchy(view, needle):
-  foundLocally = False
+def accessibilityGrepHierarchy(view, needle, found):
+  foundLocally = found
   a11yLabel = accessibilityLabel(view)
 
   #if we don't have any accessibility string - we should have some children
   if int(a11yLabel.GetValue(), 16) == 0:
     #We call private method that gives back all visible accessibility children for view
-    accessibilityElements = fb.evaluateExpression('(id)[[[UIApplication sharedApplication] keyWindow] _accessibilityElementsInContainer:0 topLevel:%s includeKB:0]' % view)
-    accessibilityElementsCount = int(fb.evaluateExpression('(int)[%s count]' % accessibilityElements))
+    accessibilityElements = fb.evaluateObjectExpression('[[[UIApplication sharedApplication] keyWindow] _accessibilityElementsInContainer:0 topLevel:%s includeKB:0]' % view)
+    accessibilityElementsCount = fb.evaluateIntegerExpression('[%s count]' % accessibilityElements)
     for index in range(0, accessibilityElementsCount):
-      subview = fb.evaluateExpression('(id)[%s objectAtIndex:%i]' % (accessibilityElements, index))
-      foundLocally |= accessibilityGrepHierarchy(subview, needle)
+      subview = fb.evaluateObjectExpression('[%s objectAtIndex:%i]' % (accessibilityElements, index))
+      foundLocally |= accessibilityGrepHierarchy(subview, needle, foundLocally)
   elif re.match(r'.*' + needle + '.*', a11yLabel.GetObjectDescription(), re.IGNORECASE):
     classDesc = objHelpers.className(view)
     print('({} {}) {}'.format(classDesc, view, a11yLabel.GetObjectDescription()))
--- a/commands/FBAccessibilityCommands.py
+++ b/commands/FBAccessibilityCommands.py
@@ -1,0 +1,108 @@
+#!/usr/bin/python
+
+# Copyright (c) 2015, Facebook, Inc.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree. An additional grant
+# of patent rights can be found in the PATENTS file in the same directory.
+
+import re
+import os
+
+import lldb
+import fblldbbase as fb
+import fblldbobjecthelpers as objHelpers
+
+# This is the key corresponding to accessibility label in _accessibilityElementsInContainer:
+ACCESSIBILITY_LABEL_KEY = 2001
+
+def lldbcommands():
+  return [
+    FBPrintAccessibilityLabels(),
+    FBFindViewByAccessibilityLabelCommand(),
+  ]
+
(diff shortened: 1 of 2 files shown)


gkassabli:
Hey, @KingOfBrian.
That's a great idea. I haven't looked into that, but I will try to add this in next iteration of command

kastiglione:
@KingOfBrian How does this pull request compare / contrast to #55?

KingOfBrian:
Much better approach, with fewer round trips by a large margin.   The different types of AX values can be added pretty easily.  forceStartAccessibilityServer is awesome, I'm gonna add that to FRY to enable AX.

I've never determined traversal choices via a nil accessiblityLabel, and I think it may cut the tree off a bit early.  A UITableView accessibilityLabel for instance is used to store the empty table view AX value, and without actually testing, I'm not sure how this would work.   There are other instances, like a navigation bar item where the accessibilityLabel is duplicated in an internal view.   It's not important to pick up that distinction, but for a debug tool, I think it'd be better to be more liberal.    And, some magic inside of _accessibilityElementsInContainer may behave differently than I think.

Anyway, this approach is definitely the correct one.  I don't have time to test it, as my AX work has passed for now.

commands/FBAccessibilityCommands.py
```diff
@@ -0,0 +1,108 @@
...
+    accessibilityElementsCount = int(fb.evaluateExpression('(int)[%s count]' % accessibilityElements))
+    for index in range(0, accessibilityElementsCount):
+      subview = fb.evaluateExpression('(id)[%s objectAtIndex:%i]' % (accessibilityElements, index))
+      foundLocally |= accessibilityGrepHierarchy(subview, needle)
+  elif re.match(r'.*' + needle + '.*', a11yLabel.GetObjectDescription(), re.IGNORECASE):
+    classDesc = objHelpers.className(view)
+    print('({} {}) {}'.format(classDesc, view, a11yLabel.GetObjectDescription()))
+
+    #First element that is found is copied to clipboard
+    if not foundLocally:
```
kastiglione: It seems this will always be true, since `foundLocally` will always be false in this `elif` branch.

commands/FBAccessibilityCommands.py
```diff
@@ -0,0 +1,108 @@
...
+
+  def description(self):
+    return 'Find the views whose accessibility labels match labelRegex and puts the address of the first result on the clipboard.'
+
+  def args(self):
+    return [ fb.FBCommandArgument(arg='labelRegex', type='string', help='The accessibility label regex to search the view hierarchy for.') ]
+
+  def run(self, arguments, options):
+    forceStartAccessibilityServer()
+    rootView = fb.evaluateExpression('(id)[[UIApplication sharedApplication] keyWindow]')
```
kastiglione: Use `evaluateObjectExpression`.

commands/FBAccessibilityCommands.py
```diff
@@ -0,0 +1,108 @@
...
+  if not fb.evaluateBooleanExpression('[UIView instancesRespondToSelector:@selector(_accessibilityElementsInContainer:)]'):
+    #Starting accessibility server is different for simulator and device
+    if fb.evaluateBooleanExpression('((NSRange)[(NSString *)[(id)[(id)[UIDevice currentDevice] model] lowercaseString] rangeOfString:@"simulator"]).location != __LONG_MAX__'):
+      lldb.debugger.HandleCommand('expr (void)[[UIApplication sharedApplication] accessibilityActivate]')
+    else:
+      lldb.debugger.HandleCommand('expr (void)[[[UIApplication sharedApplication] _accessibilityBundlePrincipalClass] _accessibilityStartServer]')
+
+def accessibilityLabel(view):
+  #using Apple private API to get real value of accessibility string for element.
+  return fb.evaluateExpressionValue('(id)[%s accessibilityAttributeValue:%i]' % (view, ACCESSIBILITY_LABEL_KEY), False)
```
kastiglione: Use `evaluateObjectExpression`.

commands/FBAccessibilityCommands.py
```diff
@@ -0,0 +1,108 @@
...
+def printAccessibilityHierarchy(view, indent = 0):
+  a11yLabel = accessibilityLabel(view)
+  classDesc = objHelpers.className(view)
+  indentString = '   | ' * indent
+
+  #if we don't have any accessibility string - we should have some children
+  if int(a11yLabel.GetValue(), 16) == 0:
+    print indentString + ('{} {}'.format(classDesc, view))
+    #We call private method that gives back all visible accessibility children for view
+    accessibilityElements = fb.evaluateExpression('(id)[[[UIApplication sharedApplication] keyWindow] _accessibilityElementsInContainer:0 topLevel:%s includeKB:0]' % view)
```
kastiglione: Use `evaluateObjectExpression`.

commands/FBAccessibilityCommands.py
```diff
@@ -0,0 +1,108 @@
...
+    accessibilityElementsCount = int(fb.evaluateExpression('(int)[%s count]' % accessibilityElements))
+    for index in range(0, accessibilityElementsCount):
+      subview = fb.evaluateExpression('(id)[%s objectAtIndex:%i]' % (accessibilityElements, index))
```
kastiglione: Use `evaluateObjectExpression`.

commands/FBAccessibilityCommands.py
```diff
@@ -0,0 +1,108 @@
...
+  foundLocally = False
+  a11yLabel = accessibilityLabel(view)
+
+  #if we don't have any accessibility string - we should have some children
+  if int(a11yLabel.GetValue(), 16) == 0:
+    #We call private method that gives back all visible accessibility children for view
+    accessibilityElements = fb.evaluateExpression('(id)[[[UIApplication sharedApplication] keyWindow] _accessibilityElementsInContainer:0 topLevel:%s includeKB:0]' % view)
+    accessibilityElementsCount = int(fb.evaluateExpression('(int)[%s count]' % accessibilityElements))
+    for index in range(0, accessibilityElementsCount):
+      subview = fb.evaluateExpression('(id)[%s objectAtIndex:%i]' % (accessibilityElements, index))
```
kastiglione: Use `evaluateObjectExpression`.

commands/FBAccessibilityCommands.py
```diff
@@ -0,0 +1,108 @@
...
+    print indentString + ('({} {}) {}'.format(classDesc, view, a11yLabel.GetObjectDescription()))
+
+def accessibilityGrepHierarchy(view, needle):
+  foundLocally = False
+  a11yLabel = accessibilityLabel(view)
+
+  #if we don't have any accessibility string - we should have some children
+  if int(a11yLabel.GetValue(), 16) == 0:
+    #We call private method that gives back all visible accessibility children for view
+    accessibilityElements = fb.evaluateExpression('(id)[[[UIApplication sharedApplication] keyWindow] _accessibilityElementsInContainer:0 topLevel:%s includeKB:0]' % view)
```
kastiglione: Use `evaluateObjectExpression`.

commands/FBAccessibilityCommands.py
```diff
@@ -0,0 +1,108 @@
...
+  a11yLabel = accessibilityLabel(view)
+  classDesc = objHelpers.className(view)
+  indentString = '   | ' * indent
+
+  #if we don't have any accessibility string - we should have some children
+  if int(a11yLabel.GetValue(), 16) == 0:
+    print indentString + ('{} {}'.format(classDesc, view))
+    #We call private method that gives back all visible accessibility children for view
+    accessibilityElements = fb.evaluateExpression('(id)[[[UIApplication sharedApplication] keyWindow] _accessibilityElementsInContainer:0 topLevel:%s includeKB:0]' % view)
+    accessibilityElementsCount = int(fb.evaluateExpression('(int)[%s count]' % accessibilityElements))
```
kastiglione: Use `evaluateIntegerExpression`.

commands/FBAccessibilityCommands.py
```diff
@@ -0,0 +1,108 @@
...
+
+def accessibilityGrepHierarchy(view, needle):
+  foundLocally = False
+  a11yLabel = accessibilityLabel(view)
+
+  #if we don't have any accessibility string - we should have some children
+  if int(a11yLabel.GetValue(), 16) == 0:
+    #We call private method that gives back all visible accessibility children for view
+    accessibilityElements = fb.evaluateExpression('(id)[[[UIApplication sharedApplication] keyWindow] _accessibilityElementsInContainer:0 topLevel:%s includeKB:0]' % view)
+    accessibilityElementsCount = int(fb.evaluateExpression('(int)[%s count]' % accessibilityElements))
```
kastiglione: Use `evaluateIntegerExpression`.

commands/FBAccessibilityCommands.py
```diff
@@ -0,0 +1,108 @@
...
+  def run(self, arguments, options):
+    forceStartAccessibilityServer()
+    rootView = fb.evaluateExpression('(id)[[UIApplication sharedApplication] keyWindow]')
+    accessibilityGrepHierarchy(rootView, arguments[0])
+
+def forceStartAccessibilityServer():
+  #We try to start accessibility server only if we don't have needed method active
+  if not fb.evaluateBooleanExpression('[UIView instancesRespondToSelector:@selector(_accessibilityElementsInContainer:)]'):
+    #Starting accessibility server is different for simulator and device
+    if fb.evaluateBooleanExpression('((NSRange)[(NSString *)[(id)[(id)[UIDevice currentDevice] model] lowercaseString] rangeOfString:@"simulator"]).location != __LONG_MAX__'):
```
kastiglione: What do you think of getting the model as a string, then do the string search in python? Might be clearer that way.

Alternatively, is there any benefit choosing which selector to call based on availability, such as `respondsToSelector:@selector(accessibilityActivate)`?

gkassabli: Unfortunately, despite the fact that accessibility behaves very differently in simulator and device - methods used are pretty much the same, so I didn't find any selector that can be used for this.
I will change search to be python-based, though

commands/FBAccessibilityCommands.py
gkassabli: You're right. Did it correct way initially, but refactored wrong way

commands/FBAccessibilityCommands.py
gkassabli: Not sure how to do use it with a string. If I try evaluateObjectExpression(...) I get string address, instead of actual string in case or evaluateExpressionValue(...).getObjectDescription(). Any suggestion on how to do it right?

kastiglione: False positive on my part. This can be left as is, apologies for the misdirection.

commands/FBAccessibilityCommands.py
kastiglione: Did you intend to correct this in 55cca664bd211f226026a51d5413350a76645e29?

commands/FBAccessibilityCommands.py
```diff
@@ -0,0 +1,108 @@
...
+  def run(self, arguments, options):
+    forceStartAccessibilityServer()
+    rootView = fb.evaluateObjectExpression('[[UIApplication sharedApplication] keyWindow]')
+    accessibilityGrepHierarchy(rootView, arguments[0], False)
+
+def forceStartAccessibilityServer():
+  #We try to start accessibility server only if we don't have needed method active
+  if not fb.evaluateBooleanExpression('[UIView instancesRespondToSelector:@selector(_accessibilityElementsInContainer:)]'):
+    #Starting accessibility server is different for simulator and device
+    if fb.evaluateExpressionValue('(id)[[UIDevice currentDevice] model]').GetObjectDescription().lower().find('simulator') >= 0:
```
kastiglione: :+1:

kastiglione:
Thanks @a2, and apologies @gkassabli for the dropped ball. I had [asked a question](https://github.com/facebook/chisel/pull/70#discussion_r23196151) and was waiting to hear back, but perhaps it wasn't seen.

gkassabli:
@kastiglione Yeah, I fixed it there. Apologies for not responding.

commands/FBAccessibilityCommands.py
gkassabli: Yes, this should work properly now

PR #72, opened 2015-01-14
add ppath command

dopcn:
Print application's 'Documents' directory path.

### DIFF
--- a/commands/FBPrintCommands.py
+++ b/commands/FBPrintCommands.py
@@ -304,7 +304,7 @@ class FBPrintKeyPath(fb.FBCommand):
 
 class FBPrintApplicationDocumentsPath(fb.FBCommand):
   def name(self):
-    return 'ppath'
+    return 'pdocspath'
 
   def description(self):
     return "Print application's 'Documents' directory path."
--- a/commands/FBPrintCommands.py
+++ b/commands/FBPrintCommands.py
@@ -29,6 +29,7 @@ def lldbcommands():
     FBPrintInternals(),
     FBPrintInstanceVariable(),
     FBPrintKeyPath(),
+    FBPrintApplicationDocumentsPath(),
   ]
 
 class FBPrintViewHierarchyCommand(fb.FBCommand):
@@ -299,3 +300,15 @@ class FBPrintKeyPath(fb.FBCommand):
       object = fb.evaluateObjectExpression(objectToMessage)
       printCommand = 'po [{} valueForKeyPath:@"{}"]'.format(object, keypath)
       lldb.debugger.HandleCommand(printCommand)
+
+
+class FBPrintApplicationDocumentsPath(fb.FBCommand):
+  def name(self):
+    return 'ppath'
+
+  def description(self):
+    return "Print application's 'Documents' directory path."
+
+  def run(self, arguments, options):
+    lldb.debugger.HandleCommand('po [NSSearchPathForDirectoriesInDomains(9, 1, YES) lastObject]')  # NSDocumentDirectory == 9 NSUserDomainMask == 1
+
--- a/commands/FBPrintCommands.py
+++ b/commands/FBPrintCommands.py
@@ -308,7 +308,19 @@ class FBPrintApplicationDocumentsPath(fb.FBCommand):
 
   def description(self):
     return "Print application's 'Documents' directory path."
+  
+  def options(self):
+    return [
+      fb.FBCommandArgument(short='-o', long='--open', arg='open', boolean=True, default=False, help='open in Finder'),
+    ]
 
   def run(self, arguments, options):
-    lldb.debugger.HandleCommand('po [NSSearchPathForDirectoriesInDomains(9, 1, YES) lastObject]')  # NSDocumentDirectory == 9 NSUserDomainMask == 1
+    if options.open:
+      path = fb.evaluateExpressionValue('(NSString *)[NSSearchPathForDirectoriesInDomains(9, 1, YES) lastObject]')
+      pathString = '{}'.format(path).split('"')[1]
+      print pathString
+      lldb.debugger.HandleCommand('script os.system("open {}")'.format(pathString))
+    else:
+      lldb.debugger.HandleCommand('po [NSSearchPathForDirectoriesInDomains(9, 1, YES) lastObject]')  # NSDocumentDirectory == 9 NSUserDomainMask == 1
+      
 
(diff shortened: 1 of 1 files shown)


commands/FBPrintCommands.py
```diff
@@ -299,3 +300,15 @@ def run(self, arguments, options):
       object = fb.evaluateObjectExpression(objectToMessage)
       printCommand = 'po [{} valueForKeyPath:@"{}"]'.format(object, keypath)
       lldb.debugger.HandleCommand(printCommand)
+
+
+class FBPrintApplicationDocumentsPath(fb.FBCommand):
+  def name(self):
+    return 'ppath'
```
arigrant: I feel like `pdocspath` might be a better command here, even if it is a bit verbose. Luckily the prefix matching will then allow `pdocs` to work as well. Thoughts?

dopcn: have thought about this before. It seems to me the word `docs path` is a little ambiguous, but use `ppath` here is just easy to type in. will change to the name suggested.

kastiglione: Thought: have `ppath` capable of printing other paths other than docs, and have it take an argument to specify which specific path. Otherwise, perhaps `ppath` can just print the app root, and then it's easy for users to append Documents or any other subpath.

commands/FBPrintCommands.py
```diff
@@ -299,3 +300,15 @@ def run(self, arguments, options):
...
+
+class FBPrintApplicationDocumentsPath(fb.FBCommand):
+  def name(self):
+    return 'pdocspath'
+
+  def description(self):
+    return "Print application's 'Documents' directory path."
+
+  def run(self, arguments, options):
+    lldb.debugger.HandleCommand('po [NSSearchPathForDirectoriesInDomains(9, 1, YES) lastObject]')  # NSDocumentDirectory == 9 NSUserDomainMask == 1
```
kastiglione: In other parts of Chisel, in addition to printing a value, we've copied it to the pasteboard. Would you mind adding that here?

kastiglione:
@dopcn What do you think about having option flags to allow the desired directory to be opened in Terminal, or maybe Finder as well? For example `ppath -t` to open in Terminal. These don't need to be added to this pull request, but they'd be nice to add sooner or later.

commands/FBPrintCommands.py
```diff
@@ -299,3 +300,27 @@ def run(self, arguments, options):
...
+    return "Print application's 'Documents' directory path."
+  
+  def options(self):
+    return [
+      fb.FBCommandArgument(short='-o', long='--open', arg='open', boolean=True, default=False, help='open in Finder'),
+    ]
+
+  def run(self, arguments, options):
+    if options.open:
+      path = fb.evaluateExpressionValue('(NSString *)[NSSearchPathForDirectoriesInDomains(9, 1, YES) lastObject]')
```
kastiglione: Can you make constants for these numbers?

kastiglione: This could use `evaluateObjectExpression()` instead of `evaluateExpressionValue()`.

commands/FBPrintCommands.py
```diff
@@ -299,3 +300,27 @@ def run(self, arguments, options):
...
+class FBPrintApplicationDocumentsPath(fb.FBCommand):
+  def name(self):
+    return 'pdocspath'
+
+  def description(self):
+    return "Print application's 'Documents' directory path."
+  
+  def options(self):
+    return [
+      fb.FBCommandArgument(short='-o', long='--open', arg='open', boolean=True, default=False, help='open in Finder'),
```
kastiglione: :+1:

commands/FBPrintCommands.py
```diff
@@ -299,3 +300,27 @@ def run(self, arguments, options):
...
+    ]
+
+  def run(self, arguments, options):
+    if options.open:
+      path = fb.evaluateExpressionValue('(NSString *)[NSSearchPathForDirectoriesInDomains(9, 1, YES) lastObject]')
+      pathString = '{}'.format(path).split('"')[1]
+      print pathString
+      lldb.debugger.HandleCommand('script os.system("open {}")'.format(pathString))
```
kastiglione: This can call `os.system()` directly, it's not necessary to call it via `HandleCommand()`.

kastiglione:
@dopcn What do you think of the comments I made?

dopcn:
@kastiglione sorry, back and working on it now...

commands/FBPrintCommands.py
dopcn: @kastiglione hi I tried `evaluateObjectExpression()` and `evaluateExpression()` but only get pointer value instead of string value so I keep using `evaluateExpressionValue()` and split the string output in the new implementation. Am I missing something?

dopcn:
**Update** copy to the pasteboard before print

commands/FBPrintCommands.py
```diff
@@ -299,3 +300,30 @@ def run(self, arguments, options):
...
+      fb.FBCommandArgument(short='-o', long='--open', arg='open', boolean=True, default=False, help='open in Finder'),
+    ]
+
+  def run(self, arguments, options):
+    # in iOS SDK NSDocumentDirectory == 9  NSUserDomainMask == 1 
+    NSDocumentDirectory = '9' 
+    NSUserDomainMask = '1'
+    path = fb.evaluateExpressionValue('(NSString*)[NSSearchPathForDirectoriesInDomains(' + NSDocumentDirectory + ', ' + NSUserDomainMask + ', YES) lastObject]')
+    pathString = '{}'.format(path).split('"')[1]
+    cmd = 'echo {} | tr -d "\n" | pbcopy'.format(pathString)
```
kastiglione: This could be `echo -n` to avoid the `tr` piece.

kastiglione:
Thanks @dopcn, I'm happy to accept this. Before that, can you merge from `master` because there are now conflicts.

commands/FBPrintCommands.py
dopcn: I don't know why `echo -n` just doesn't work as `os.system()`'s parameter. The `-n` will be taken as part of pathString and copied into pasteboard

kastiglione: Seems `/bin/sh` doesn't handle `-n` like bash and zsh do. Oh well. Thanks for trying.

kastiglione:
Thanks @dopcn!

PR #76, opened 2015-02-20
Fixed incorrect usage of global foundElement causing crashes in some searches

gkassabli:


commands/FBAccessibilityCommands.py
```diff
@@ -86,6 +86,7 @@ def printAccessibilityHierarchy(view, indent = 0):
     print indentString + ('({} {}) {}'.format(classDesc, view, a11yLabel.GetObjectDescription()))
 
 def accessibilityGrepHierarchy(view, needle):
+  global foundElement
```
kastiglione: Can these methods be made instance methods, and then `foundElement` made into an attribute? Or does this value need to persist from one chisel command invocation to the next?

kastiglione:
Thanks!

PR #83, opened 2015-03-23
Add pdata command to print content of NSData object.

bartoszj:
`pdata` prints content of NSData object. By default it is using UTF-8 encoding, but other encodings like UTF-16/UTF-32, ISO-8859-1/2, Windows CP-1250/1/2/3/4 are also supported.

Example usage:
```
pdata self.data
pdata -e utf16 self.data
```


kastiglione:
Nice, thank you!

I trust you that all the encoding values are right :)

PR #87, opened 2015-04-08
It is not necessary

apoorvcode:


kastiglione:
Thanks for the pull request @apoorvcode. Why is this no longer necessary? It was added in #69 because we have some commands that have conditional behavior based on the architecture.

kastiglione:
Thanks for the submission. Let me know if I've missed anything about why this isn't needed.

PR #90, opened 2015-04-29
Don't skip prologue for `bmessage` breakpoints

### LINKED ISSUE #89: bmessage: Consult skip-prologue setting on x86
If the `target.skip-prologue` setting is set to `true`, then [`functionPreambleExpressionForSelf()` will return an invalid expression](https://github.com/facebook/chisel/blob/6c1cfcd4010100eb54b1802fab27e4d1726aed5e/fblldbobjcruntimehelpers.py#L50-L51). When the prologue is skipped, the expression should be `*(id*)($ebp+8)`.

cc @rohan-mehta

kastiglione:
The x86 expression returned from `functionPreambleExpressionForSelf()` expects the breakpoint to be stopped before the prologue. The other architectures should work before or after prologue.

Resolves #89

### DIFF
--- a/commands/FBDebugCommands.py
+++ b/commands/FBDebugCommands.py
@@ -158,10 +158,10 @@ class FBMethodBreakpointCommand(fb.FBCommand):
     print 'Setting a breakpoint at {} with condition {}'.format(breakpointFullName, breakpointCondition)
 
     if category:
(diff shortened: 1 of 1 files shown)


kastiglione:
@mattjgalloway Thoughts on this?

mattjgalloway:
@kastiglione LGTM. Ship it!

kastiglione:
:ok_hand: thanks

Issue #88, opened 2015-04-13
'vs' command broken?


palcalde:
XCode prints this, anyone knows how to fix it?

Traceback (most recent call last):
  File "/usr/local/Cellar/chisel/1.2.0/libexec/fblldb.py", line 81, in runCommand
    command.run(args, options)
  File "/usr/local/Cellar/chisel/1.2.0/libexec/commands/FBFlickerCommands.py", line 60, in run
    walker = FlickerWalker(object)
  File "/usr/local/Cellar/chisel/1.2.0/libexec/commands/FBFlickerCommands.py", line 67, in __init__
    self.handler = inputHelpers.FBInputHandler(lldb.debugger, self.inputCallback)
  File "/usr/local/Cellar/chisel/1.2.0/libexec/fblldbinputhelpers.py", line 16, in __init__
    self.inputReader = lldb.SBInputReader()
AttributeError: 'module' object has no attribute 'SBInputReader'

kastiglione:
Thanks for the report! I'll look into this and get back to you, hopefully soon :stuck_out_tongue_winking_eye:

palcalde:
Thank you! I used to use this command a lot to jump between siblings and superviews. It was great. And a lot faster than the built in tree hierarchy view in the debugger :-)

kastiglione:
@palcalde Which version of Xcode are you using? With Xcode 6.2 / lldb-320.4.160, the problem is that the `SBInputReader` class is not exposed through the `lldb` module. I haven't tried Xcode 6.3, hopefully the missing class is a temporary regression.

palcalde:
@kastiglione I'm using 6.3, tho this command has been broken since last 3/4 months. I think since XCode 6.0. I've created a #92  that fixes it, instead of using SBInputReader which, as you mention, is not accesible anymore I used the standard stdin. I tested it and seems to work fine. Let me know what you think.

kastiglione:
Fixed by #92

PR #92, opened 2015-05-04
Fixed ‘vs’ command by:

palcalde:
Fixed ‘vs’ command by (Issue https://github.com/facebook/chisel/issues/88#issuecomment-98440239):

- Stop using FBInputHandler. This class uses SBInputReader which no
longer exist in the lldb version that comes bundled with XCode (that’s
why the command was broken), even tho it does exist in the official
lldb documentation
(http://lldb.llvm.org/python_reference/lldb.SBInputReader-class.html).

- Now the script uses the standard stdin library to read from terminal.

- Removed the ‘Flickering’ of the selected view, instead we now use
mask which is a lot more useful visually.

NOTE: As mentioned this fix doesn't use the class FBInputHandler at all to avoid SBInputReader, but if we have that class back in next versions of XCode we can always revert this fix-commit.

### DIFF
--- a/commands/FBFlickerCommands.py
+++ b/commands/FBFlickerCommands.py
@@ -115,7 +115,7 @@ class FlickerWalker:
 
       lldb.debugger.HandleCommand('po [(id)' + oldView + ' ' + recusionName + ']')
     else:
-      print '\nChisel - VS Mode: I really have no idea what you meant by \'' + input + '\'... =\\\n'
+      print '\nI really have no idea what you meant by \'' + input + '\'... =\\\n'
 
   def setCurrentView(self, view, oldView=None):
     if view:
--- a/commands/FBFlickerCommands.py
+++ b/commands/FBFlickerCommands.py
@@ -63,17 +63,17 @@ class FBViewSearchCommand(fb.FBCommand):
 
 class FlickerWalker:
   def __init__(self, startView):
-    self.setCurrentView(startView, None)
+    self.setCurrentView(startView)
 
   def run(self):
     self.keepRunning = True
     initialAsync = lldb.debugger.GetAsync()
-    lldb.debugger.SetAsync(True) #needed so XCode doesn't hang if tap on Continue while lldb is waiting for user input in 'vs' mode
+    lldb.debugger.SetAsync(True) #Needed so XCode doesn't hang if tap on Continue while lldb is waiting for user input in 'vs' mode
     while self.keepRunning:
       charRead = sys.stdin.readline().rstrip("\n")
       self.inputCallback(charRead)
     else:
-      lldb.debugger.SetAsync(initialAsync) #restore to init value
+      lldb.debugger.SetAsync(initialAsync)
 
   def inputCallback(self, input):
     oldView = self.currentView
--- a/commands/FBFlickerCommands.py
+++ b/commands/FBFlickerCommands.py
@@ -9,6 +9,7 @@
 
 import os
 import time
+import sys
 
 import lldb
 import fblldbbase as fb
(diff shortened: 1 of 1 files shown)


commands/FBFlickerCommands.py
```diff
@@ -116,13 +112,16 @@ def inputCallback(self, input):
 
       lldb.debugger.HandleCommand('po [(id)' + oldView + ' ' + recusionName + ']')
     else:
-      print '\nI really have no idea what you meant by \'' + input + '\'... =\\\n'
+      print '\nChisel - VS Mode: I really have no idea what you meant by \'' + input + '\'... =\\\n'
 
     viewHelpers.setViewHidden(oldView, False)
 
-  def setCurrentView(self, view):
+  def setCurrentView(self, view, oldView):
```
kastiglione: Default `oldView` to `None`?

commands/FBFlickerCommands.py
```diff
@@ -62,20 +63,14 @@ def run(self, arguments, options):
...
-    while self.handler.isValid():
-      self.flicker()
-
-  def flicker(self):
-    viewHelpers.setViewHidden(self.currentView, True)
-    time.sleep(0.1)
-    viewHelpers.setViewHidden(self.currentView, False)
-    time.sleep(0.3)
+    self.keepRunning = True
+    lldb.debugger.SetAsync (True)
```
kastiglione: Can you comment on why this is necessary? And, remove the space after `SetAsync`?

kastiglione:
@palcalde Thanks!

I've left a couple comments.

Also, can we make the mask a flag? (You can see examples of how to do flags in other chisel commands.) I'd prefer not to arbitrarily change the command from flickering to masking. Ideally, this could also use a border instead of flicker or mask, but I'm not asking you to go that far :stuck_out_tongue_winking_eye: .

palcalde:
@kastiglione thanks for reviewing. I will add a comment about the async mode. The reason is because if you are on lldb console and type vs, the script blocks the thread waiting for user input. That is expected, however, if you type on continue in the debugger XCode enables the simulator again but lldb becomes absolutely blocked. You can't stop the execution or anything like that. The only thing you can do is quit simulator to kill the lldb process. 

I found that if you run in async mode, lldb runs in a diff thread so it doesn't become blocked when you continue debugger in vs mode. You can stop again and keep using it normally. 

About the border instead of masking, I will investigate and try to see if I can put it in the open PR.

kastiglione:
> About the border instead of masking, I will investigate and try to see if I can put it in the open PR.

To be clear, it's not important to add border support. I'm only concerned about keeping flicker support. I'd rather not remove flicker support, but if we want to augment it with mask support, :+1:.

For async mode, are there consequences to leaving it in async mode? Or should it be returned to off after `vs` completes?

palcalde:
@kastiglione The only reason I moved flickering to masking is because the new implementation doesn't use SBInputReader, it uses stdin which blocks the thread waiting for user input. SBInputReader was a separated buffer so the previous implementation did a while hiding/showing the view every 0.1 sec. 

Since adding an async implementation with stdin in python is not trivial (it would require a Queue a Thread and a couple of methods) I just thought it wasn't worth it and just went for a simple mask behavior. That is the only reason why I change it :)

I added one more commit fixing the issues you mentioned.

commands/FBFlickerCommands.py
```diff
@@ -62,20 +63,17 @@ def run(self, arguments, options):
 
 class FlickerWalker:
   def __init__(self, startView):
-    self.setCurrentView(startView)
-
-    self.handler = inputHelpers.FBInputHandler(lldb.debugger, self.inputCallback)
-    self.handler.start()
+    self.setCurrentView(startView, None)
```
kastiglione: Now that `oldView` defaults to `None`, the `None` here can be removed.

kastiglione:
Thanks for the explanation. Let's do it. A masking `vs` is better than a broken `vs`.

I have one more comment, otherwise looks good and thank you so much for the pull request!

palcalde:
Sure, no worries it was fun dealing with lldb and Python a little bit! :+1: 
I will update the pull request with your comment.

palcalde:
Pull Request updated.

commands/FBFlickerCommands.py
```diff
@@ -116,13 +115,14 @@ def inputCallback(self, input):
 
       lldb.debugger.HandleCommand('po [(id)' + oldView + ' ' + recusionName + ']')
     else:
-      print '\nI really have no idea what you meant by \'' + input + '\'... =\\\n'
-
-    viewHelpers.setViewHidden(oldView, False)
+      print '\nChisel - VS Mode: I really have no idea what you meant by \'' + input + '\'... =\\\n'
```
kastiglione: Just noticed this. We don't really do this anywhere else with any output. Do you think it's necessary to mention Chisel and VS by name?

palcalde:
The only reason why I put that in the message is because is the only command in Chisel (that I know) that stops lldb waiting for user input. 

I thought it could be useful to remind people they are in vs mode. 

I can reverse it if you feel more comfortable. Just let me know!

palcalde:
@kastiglione reverted message so is consistent with the rest of commands.

kastiglione:
:fireworks:

Issue #59, opened 2014-12-08
Add target/action commands


kastiglione:
As described in the [Dancing in the Debugger — A Waltz with LLDB](http://www.objc.io/issue-19/lldb-debugging.html), it's straight forward to manually determine the targets/actions of a control. However, what was new to me in that article is the ability to use a `0` value `UIControlEvents`, as a form of a wildcard.

I think there might be room for two commands: `pactions`, and `bactions`.

cc @arigrant

kastiglione:
Resolved by #95

PR #95, opened 2015-05-24
Add `pactions` command to print the target/actions of a UIControl

kastiglione:
I've been meaning to add a `pactions` command for a while. It's pretty basic. It would be nice to print out which control event the action observes, but this covers probably 90+% of the cases. In #59, I mentioned it might be nice to have a `bactions` command too, but I'm thinking that's a wait and see enhancement.

For a view controller, an example of the output is:

`<XXXViewController: 0xdddddddd>: didTapButton:`

and if there are multiple actions:

`<XXXViewController: 0xdddddddd>: didTapButton:, didTouchButton:`

cc @arigrant

### DIFF
--- a/commands/FBPrintCommands.py
+++ b/commands/FBPrintCommands.py
@@ -30,6 +30,7 @@ def lldbcommands():
     FBPrintInstanceVariable(),
     FBPrintKeyPath(),
     FBPrintData(),
+    FBPrintTargetActions(),
   ]
 
 class FBPrintViewHierarchyCommand(fb.FBCommand):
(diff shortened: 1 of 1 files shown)


commands/FBPrintCommands.py
```diff
@@ -371,3 +372,28 @@ def run(self, arguments, option):
...
+    control = arguments[0]
+    targets = fb.evaluateObjectExpression('[[{control} allTargets] allObjects]'.format(control=control))
+    targetCount = fb.evaluateIntegerExpression('[{targets} count]'.format(targets=targets))
+
+    for index in range(0, targetCount):
+      target = fb.evaluateObjectExpression('[{targets} objectAtIndex:{index}]'.format(targets=targets, index=index))
+      actions = fb.evaluateObjectExpression('[{control} actionsForTarget:{target} forControlEvent:0]'.format(control=control, target=target))
+
+      targetDescription = fb.evaluateExpressionValue('(id){target}'.format(target=target)).GetObjectDescription()
+      actionsDescription = fb.evaluateExpressionValue('(id)[{actions} componentsJoinedByString:@","]'.format(actions=actions)).GetObjectDescription()
```
arigrant: What does this look like when it prints?

kastiglione: Ah yes, I meant to include an example output, thanks for asking!

For a view controller, something like:

`<XXXViewController: 0xdddddddd>: didTapButton:`

The output would be messier if the target provides a verbose `-description` method. Perhaps this shouldn't rely on the built in `-description` and should instead manually format the `<Class: 0xdddddddd>` string.

PR #97, opened 2015-06-01
Exception fixes in functionPreambleExpression*

kastiglione:
Minor fixes in `functionPreambleExpressionForObjectParameterAtIndex`. Fixes an exception message for x86-64, and an exception condition for arm32.

PR #98, opened 2015-06-01
Use register aliases

kastiglione:
:warning: I haven't tested this on arm :warning: 

Use lldb's register aliases instead of specific architecture register names. The main reason for this change is that some commands, like `bmessage`, will print out the results that include the output of these functions. I think it would be beneficial to surface the aliases to users, to spread their existence and availability.

Todo:

- [ ] Test on arm32
- [ ] Test on arm64

Depends on #97

### DIFF
--- a/fblldbobjcruntimehelpers.py
+++ b/fblldbobjcruntimehelpers.py
@@ -45,19 +45,10 @@ def currentArch():
   return arch
 
 def functionPreambleExpressionForSelf():
-  import re
-
-  arch = currentArch()
-  expressionForSelf = None
-  if arch == 'i386':
-    expressionForSelf = '*(id*)($esp+4)'
-  elif arch == 'x86_64':
-    expressionForSelf = '(id)$rdi'
-  elif arch == 'arm64':
-    expressionForSelf = '(id)$x0'
-  elif re.match(r'^armv.*$', arch):
-    expressionForSelf = '(id)$r0'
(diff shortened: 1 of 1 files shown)


mattjgalloway:
We should ship this. @kastiglione did you check it works on ARM yet?

kastiglione:
Thanks for the ping, will put this on my todo list.

fblldbobjcruntimehelpers.py
```diff
@@ -45,39 +45,22 @@ def currentArch():
...
-    expresssion = '(id)$r' + str(parameterIndex + 2)
-  return expresssion
+    return '*(id*)($esp+{offset})'.format(offset=12 + parameterIndex * 4)
+  elif arch == 'x86_64' and parameterIndex > 3:
+    raise Exception("Current implementation can not return object at index greater than 3 for x86_64")
+  elif arch == 'arm64' and parameterIndex > 5:
+    raise Exception("Current implementation can not return object at index greater than 5 for arm64")
+  elif re.match(r'^armv.*$', arch) and parameterIndex > 1:
+    raise Exception("Current implementation can not return object at index greater than 1 for arm32")
+  return '(id)$arg{n}'.format(n=parameterIndex + 2)
```
kastiglione: Note to self, I think this should be `parameterIndex + 3` because the `$argN` alias starts at 1.

PR #99, opened 2015-06-02
Prevent overly eager regex match in `fv` command

kastiglione:
The regex used in `fv` is overly eager about how it matches the class name. This change replaces a `.*` with `\S*`. Without this, a formatted object embedded in a view's description could be matched as the address of the view. For example:

`<ABCOneTwoThreeView: 0x25f9f180; frame = (12 606; 401 401); image = <UIImage: 0x215e9a80; …>>`

Previously, the `fv` command would match the second address (`0x215e9a80`). This change will capture the first address in this case.


mattjgalloway:
LGTM. I'll merge it.

kastiglione:
Thanks!

PR #104, opened 2015-08-04
Add ability to visualize UIColor, CIColor, and CGColorRef.

### LINKED ISSUE #103: Ability to visualize colors
If I run `visualize someColor`, I get:

```
UIDeviceRGBColor isn't supported. You can visualize UIImage, CGImageRef, UIView, CALayer or NSData.
```

Chisel should generate a square or circular image of the color, optionally with some RGB/hex info included in the image, and open the image in Preview.

ZevEisenberg:
Looks like this:

screen shot 2015-08-04 at 9 57 14 am

We could add more fancy visualization, including rendering text of what the constituent colors are, like Apple does, but I figured it would be a good idea to start simple.

Fixes #103

### DIFF
--- a/commands/FBVisualizationCommands.py
+++ b/commands/FBVisualizationCommands.py
@@ -57,6 +57,50 @@ def _showImage(commandForImage):
     imgFile.close()
     os.system('open ' + imagePath)
 
+def _colorIsCGColorRef(color):
+  color = '(CGColorRef)(' + color + ')'
+
+  frame = lldb.debugger.GetSelectedTarget().GetProcess().GetSelectedThread().GetSelectedFrame()
+  result = frame.EvaluateExpression('(unsigned long)CFGetTypeID({color}) == (unsigned long)CGColorGetTypeID()'.format(color=color))
+
+  if result.GetError() is not None and str(result.GetError()) != 'success':
+    print "got error: {}".format(result)
+    return 0
+  else:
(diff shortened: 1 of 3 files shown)


commands/FBVisualizationCommands.py
```diff
@@ -57,6 +57,50 @@ def _showImage(commandForImage):
...
+
+  if result.GetError() is not None and str(result.GetError()) != 'success':
+    print "got error: {}".format(result)
+    return 0
+  else:
+    isCFColor = result.GetValueAsUnsigned() != 0
+    if isCFColor:
+      return 1
+    else:
+      return 0
```
kastiglione: What's the reason for using `0`/`1` instead of `False`/`True`?

kastiglione:
Looks good, thank you!

commands/FBVisualizationCommands.py
ZevEisenberg: Copied the pattern from elsewhere in the script. Happy to change it.

ZevEisenberg: Cone to think of it, the `!= 0` is already returning a boolean. We can just return that.

ZevEisenberg:
@kastiglione updated return statements

kastiglione:
@ZevEisenberg Thank you! If you have any other ideas, would love to get more pull requests or issues as well.

PR #109, opened 2015-09-08
pcurl command

VTopoliuk:
Added ability to convert NSURLReuqest to curl command.
Example:
```Objective-C
  NSMutableURLRequest *request = [[NSMutableURLRequest alloc] initWithURL:[NSURL URLWithString:@"https://google.com"]];
  request.timeoutInterval = 30;
  request.HTTPMethod = @"POST";
  request.allHTTPHeaderFields = @{@"Header1" : @"test-1", @"Header2" : @"some test sttring"};
  request.HTTPBody = [@"sample data" dataUsingEncoding:NSUTF8StringEncoding];
```
```
curl -X POST --connect-timeout 30 -H "Header2: some test sttring" -H "Header1: test-1" --data-binary @"/tmp/curl_data_463408775.11118102" "https://google.com"
```
Also you can embed data into terminal command with --embed-data parameter, result:
```
echo "c2FtcGxlIGRhdGE=" | base64 -D -o "/tmp/curl_data_463487823.35579801"; curl -X POST --connect-timeout 30 -H "Header2: some test sttring" -H "Header1: test-1" --data-binary @"/tmp/curl_data_463487823.35579801" "https://google.com"
```

### DIFF
--- a/commands/FBPrintCommands.py
+++ b/commands/FBPrintCommands.py
@@ -458,18 +458,23 @@ class FBPrintAsCurl(fb.FBCommand):
   def description(self):
     return 'Print the NSURLRequest (HTTP) as curl command.'
 
+  def options(self):
+    return [
+      fb.FBCommandArgument(short='-p', long='--portable', arg='portable', boolean=True, default=False, help='Embed request data as base64.'),
+    ]
+
   def args(self):
     return [ fb.FBCommandArgument(arg='request', type='NSURLRequest*/NSMutableURLRequest*', help='The request to convert to the curl command.') ]
 
   def run(self, arguments, options):
     request = arguments[0]
     HTTPHeaderSring = ''
-    HTTPMethod = fb.evaluateExpressionValue('(id)[{request} HTTPMethod]'.format(request=request)).GetObjectDescription()
-    URL = fb.evaluateExpressionValue('(id)[{request} URL]'.format(request=request)).GetObjectDescription()
-    timeout = fb.evaluateExpression('(NSTimeInterval)[{request} timeoutInterval]'.format(request=request))
-    HTTPHeaders = fb.evaluateObjectExpression('(id)[{request} allHTTPHeaderFields]'.format(request=request))
-    HTTPHeadersCount = fb.evaluateIntegerExpression('[{HTTPHeaders} count]'.format(HTTPHeaders=HTTPHeaders)) 
-    allHTTPKeys = fb.evaluateObjectExpression('[{HTTPHeaders} allKeys]'.format(HTTPHeaders=HTTPHeaders))
+    HTTPMethod = fb.evaluateExpressionValue('(id)[{} HTTPMethod]'.format(request)).GetObjectDescription()
+    URL = fb.evaluateExpressionValue('(id)[{} URL]'.format(request)).GetObjectDescription()
+    timeout = fb.evaluateExpression('(NSTimeInterval)[{} timeoutInterval]'.format(request))
+    HTTPHeaders = fb.evaluateObjectExpression('(id)[{} allHTTPHeaderFields]'.format(request))
+    HTTPHeadersCount = fb.evaluateIntegerExpression('[{} count]'.format(HTTPHeaders)) 
+    allHTTPKeys = fb.evaluateObjectExpression('[{} allKeys]'.format(HTTPHeaders))
     for index in range(0, HTTPHeadersCount):
         key = fb.evaluateObjectExpression('[{} objectAtIndex:{}]'.format(allHTTPKeys, index))
         keyDescription = fb.evaluateExpressionValue('(id){}'.format(key)).GetObjectDescription()
@@ -477,13 +482,24 @@ class FBPrintAsCurl(fb.FBCommand):
         if len(HTTPHeaderSring) > 0:
             HTTPHeaderSring += ' '
         HTTPHeaderSring += '-H "{}: {}"'.format(keyDescription, value)
-    HTTPData = fb.evaluateObjectExpression('[{request} HTTPBody]'.format(request=request))
+    HTTPData = fb.evaluateObjectExpression('[{} HTTPBody]'.format(request))
     dataFile = None
+    dataAsString = None
     if fb.evaluateIntegerExpression('[{} length]'.format(HTTPData)) > 0:
         dataFile = '/tmp/curl_data_{}'.format(fb.evaluateExpression('(NSTimeInterval)[NSDate timeIntervalSinceReferenceDate]'))
-        fb.evaluateExpression('(BOOL)[{} writeToFile:@"{}" atomically:NO]'.format(HTTPData, dataFile))
-                
-    commandString = 'curl -X {} --connect-timeout {}'.format(HTTPMethod, timeout)
+        if options.portable:
+          if fb.evaluateIntegerExpression('[{} respondsToSelector:@selector(base64EncodedStringWithOptions:)]'.format(HTTPData)):
+            dataAsString = fb.evaluateExpressionValue('(id)[(id){} base64EncodedStringWithOptions:0]'.format(HTTPData)).GetObjectDescription()
+        elif not runtimeHelpers.isIOSDevice():
+          fb.evaluateExpression('(BOOL)[{} writeToFile:@"{}" atomically:NO]'.format(HTTPData, dataFile))
+        else:
+          print 'HTTPBody data for iOS Device is supported only with "--portable" flag'
+          return False
+
+    commandString = ''
+    if dataAsString is not None and len(dataAsString) > 0:
+      commandString += 'echo "{}" | base64 -D -o "{}"; '.format(dataAsString, dataFile)
+    commandString += 'curl -X {} --connect-timeout {}'.format(HTTPMethod, timeout)
     if len(HTTPHeaderSring) > 0:
         commandString += ' ' + HTTPHeaderSring
     if dataFile is not None:
--- a/commands/FBPrintCommands.py
+++ b/commands/FBPrintCommands.py
@@ -32,10 +32,7 @@ def lldbcommands():
     FBPrintApplicationDocumentsPath(),
     FBPrintData(),
     FBPrintTargetActions(),
-<<<<<<< HEAD
     FBPrintJSON(),
-=======
->>>>>>> origin/pcurl
     FBPrintAsCurl(),
   ]
 
@@ -430,7 +427,6 @@ class FBPrintTargetActions(fb.FBCommand):
 
       print '{target}: {actions}'.format(target=targetDescription, actions=actionsDescription)
 
-<<<<<<< HEAD
 class FBPrintJSON(fb.FBCommand):
     
   def name(self):
@@ -455,8 +451,6 @@ class FBPrintJSON(fb.FBCommand):
     
     print jsonString
     
-=======
->>>>>>> origin/pcurl
 class FBPrintAsCurl(fb.FBCommand):
   def name(self):
     return 'pcurl'
--- a/commands/FBPrintCommands.py
+++ b/commands/FBPrintCommands.py
@@ -466,6 +466,9 @@ class FBPrintAsCurl(fb.FBCommand):
   def args(self):
     return [ fb.FBCommandArgument(arg='request', type='NSURLRequest*/NSMutableURLRequest*', help='The request to convert to the curl command.') ]
 
+  def generateTmpFilePath(self):
+    return '/tmp/curl_data_{}'.format(fb.evaluateExpression('(NSTimeInterval)[NSDate timeIntervalSinceReferenceDate]'))
+
   def run(self, arguments, options):
     request = arguments[0]
     HTTPHeaderSring = ''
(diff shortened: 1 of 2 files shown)


commands/FBPrintCommands.py
```diff
@@ -424,3 +425,44 @@ def run(self, arguments, options):
...
+  def description(self):
+    return 'Print the NSURLRequest (HTTP) as curl command.'
+
+  def args(self):
+    return [ fb.FBCommandArgument(arg='request', type='NSURLRequest*/NSMutableURLRequest*', help='The request to convert to the curl command.') ]
+
+  def run(self, arguments, options):
+    request = arguments[0]
+    HTTPHeaderSring = ''
+    HTTPMethod = fb.evaluateExpressionValue('(id)[{request} HTTPMethod]'.format(request=request)).GetObjectDescription()
```
mattjgalloway: I would prefer:

```
HTTPMethod = fb.evaluateExpressionValue("(id)[%s HTTPMethod]" % (request)).GetObjectDescription()
```

But then again, I ain't no Python guru.

mattjgalloway:
Overall LGTM. @kastiglione - any thoughts before we merge?

commands/FBPrintCommands.py
```diff
@@ -424,3 +425,44 @@ def run(self, arguments, options):
...
+        keyDescription = fb.evaluateExpressionValue('(id){}'.format(key)).GetObjectDescription()
+        value = fb.evaluateExpressionValue('(id)[(id){} objectForKey:{}]'.format(HTTPHeaders, key)).GetObjectDescription()
+        if len(HTTPHeaderSring) > 0:
+            HTTPHeaderSring += ' '
+        HTTPHeaderSring += '-H "{}: {}"'.format(keyDescription, value)
+    HTTPData = fb.evaluateObjectExpression('[{request} HTTPBody]'.format(request=request))
+    dataFile = None
+    if fb.evaluateIntegerExpression('[{} length]'.format(HTTPData)) > 0:
+        dataFile = '/tmp/curl_data_{}'.format(fb.evaluateExpression('(NSTimeInterval)[NSDate timeIntervalSinceReferenceDate]'))
+        fb.evaluateExpression('(BOOL)[{} writeToFile:@"{}" atomically:NO]'.format(HTTPData, dataFile))
```
mattjgalloway: Presumably this only works on the simulator then? Could we base64 this maybe and throw it into curl as a base64 encoded string?

mattjgalloway:
Overall LGTM, sans the comments. @kastiglione - any thoughts before we merge?

arigrant:
Looks great to me! I'll let @kastiglione take a peek as well though.

commands/FBPrintCommands.py
VTopoliuk: Yeah, it's a bottleneck for now. I can't find the way to pass raw data to curl.

commands/FBPrintCommands.py
kastiglione: IANAPG either, but I believe `format` is preferred over `%`. However, `format` does not require named args, so a middle ground is:

```python
HTTPMethod = fb.evaluateExpressionValue('(id)[{} HTTPMethod]'.format(request)).GetObjectDescription()
```

commands/FBPrintCommands.py
kastiglione: Let's spit out an error when attempted on device.

kastiglione:
Great addition, thanks @VTopoliuk!

I think this just needs to error when run on device and is good to go.

VTopoliuk:
I added --portable parameter to embed data into the final command. So now iOS Device is supported, but script will fire error if executed without --portable on device and request has body data.

commands/FBPrintCommands.py
```diff
@@ -424,3 +425,60 @@ def run(self, arguments, options):
...
+class FBPrintAsCurl(fb.FBCommand):
+  def name(self):
+    return 'pcurl'
+
+  def description(self):
+    return 'Print the NSURLRequest (HTTP) as curl command.'
+
+  def options(self):
+    return [
+      fb.FBCommandArgument(short='-p', long='--portable', arg='portable', boolean=True, default=False, help='Embed request data as base64.'),
```
mattjgalloway: Could this be `useBase64Data` or something? `portable` seems like a strange name for this parameter.

Or, just always do base64 data?

commands/FBPrintCommands.py
```diff
@@ -424,3 +425,60 @@ def run(self, arguments, options):
...
+        if len(HTTPHeaderSring) > 0:
+            HTTPHeaderSring += ' '
+        HTTPHeaderSring += '-H "{}: {}"'.format(keyDescription, value)
+    HTTPData = fb.evaluateObjectExpression('[{} HTTPBody]'.format(request))
+    dataFile = None
+    dataAsString = None
+    if fb.evaluateIntegerExpression('[{} length]'.format(HTTPData)) > 0:
+        dataFile = '/tmp/curl_data_{}'.format(fb.evaluateExpression('(NSTimeInterval)[NSDate timeIntervalSinceReferenceDate]'))
+        if options.portable:
+          if fb.evaluateIntegerExpression('[{} respondsToSelector:@selector(base64EncodedStringWithOptions:)]'.format(HTTPData)):
```
mattjgalloway: Probably no need to do this, since it's available in iOS 7 and I don't think we support <iOS 7. Although I can't remember for sure. @kastiglione?

commands/FBPrintCommands.py
```diff
@@ -424,3 +425,60 @@ def run(self, arguments, options):
...
+        keyDescription = fb.evaluateExpressionValue('(id){}'.format(key)).GetObjectDescription()
+        value = fb.evaluateExpressionValue('(id)[(id){} objectForKey:{}]'.format(HTTPHeaders, key)).GetObjectDescription()
+        if len(HTTPHeaderSring) > 0:
+            HTTPHeaderSring += ' '
+        HTTPHeaderSring += '-H "{}: {}"'.format(keyDescription, value)
+    HTTPData = fb.evaluateObjectExpression('[{} HTTPBody]'.format(request))
+    dataFile = None
+    dataAsString = None
+    if fb.evaluateIntegerExpression('[{} length]'.format(HTTPData)) > 0:
+        dataFile = '/tmp/curl_data_{}'.format(fb.evaluateExpression('(NSTimeInterval)[NSDate timeIntervalSinceReferenceDate]'))
```
mattjgalloway: Move this inside the `elif not runtimeHelpers.isIOSDevice():`.

commands/FBPrintCommands.py
VTopoliuk: @mattjgalloway actually 'portable' in this case means portable command or "you can run this command on other computers, not current one only"
http://dictionary.reference.com/browse/portable

base64 looks ugly and contains overhead on every run, not sure always output it is a good option.

mattjgalloway: I don't get how 'portable' makes sense here still. It sort of makes sense, but not intuitive to me.

Since you're copy-pasting the curl command anyway, I don't see why it matters that the base64 is ugly.

VTopoliuk: In my real-life test base64 is longer than one screen

mattjgalloway: Ah ok. Big amounts of data!

I still do think that `portable` is a bad name. Maybe even it should be `dataStyle` and then options are `tmp` or `base64` and it defaults to the one which makes most sense for the current platform.

mattjgalloway: Happy to see what @kastiglione / @arigrant think too.

kastiglione: "portable" is definitely opaque to me. I'm good with `--base64`. Other suggestions: `--inline(-data)`, `--embed(-data)`…

commands/FBPrintCommands.py
kastiglione: I'm not sure if we state an official iOS version support. If we don't, this is as good a time to start. So yeah, let's say iOS 7+.

VTopoliuk: This command works on OS X as well, but this method available only for 10.9+

VTopoliuk:
I updated parameter name

mattjgalloway:
@VTopoliuk How about the nit? Can you fix that and then I think we're good to go!

commands/FBPrintCommands.py
```diff
@@ -424,3 +425,60 @@ def run(self, arguments, options):
...
+          return False
+
+    commandString = ''
+    if dataAsString is not None and len(dataAsString) > 0:
+      commandString += 'echo "{}" | base64 -D -o "{}"; '.format(dataAsString, dataFile)
+    commandString += 'curl -X {} --connect-timeout {}'.format(HTTPMethod, timeout)
+    if len(HTTPHeaderSring) > 0:
+        commandString += ' ' + HTTPHeaderSring
+    if dataFile is not None:
+        commandString += ' --data-binary @"{}"'.format(dataFile)
```
VTopoliuk: @mattjgalloway actually same file are used in two cases:
1. To embed data
2. To write data to file

If I will place name generation under the `not runtimeHelpers.isIOSDevice():` condition, then it will be not available here with `--embed-data` option.

mattjgalloway: Oh whoops sorry!

VTopoliuk: Should we go?

VTopoliuk: @kastiglione do you think it's ready to merge?

commands/FBPrintCommands.py
```diff
@@ -424,3 +425,60 @@ def run(self, arguments, options):
...
+    dataAsString = None
+    if fb.evaluateIntegerExpression('[{} length]'.format(HTTPData)) > 0:
+        dataFile = '/tmp/curl_data_{}'.format(fb.evaluateExpression('(NSTimeInterval)[NSDate timeIntervalSinceReferenceDate]'))
+        if options.embed:
+          if fb.evaluateIntegerExpression('[{} respondsToSelector:@selector(base64EncodedStringWithOptions:)]'.format(HTTPData)):
+            dataAsString = fb.evaluateExpressionValue('(id)[(id){} base64EncodedStringWithOptions:0]'.format(HTTPData)).GetObjectDescription()
+        elif not runtimeHelpers.isIOSDevice():
+          fb.evaluateExpression('(BOOL)[{} writeToFile:@"{}" atomically:NO]'.format(HTTPData, dataFile))
+        else:
+          print 'HTTPBody data for iOS Device is supported only with "--portable" flag'
```
kastiglione: s/`--portable`/`--embed-data`/

commands/FBPrintCommands.py
kastiglione: It looks like this line should be moved into the `elif not runtimeHelpers.isIOSDevice():` branch.

commands/FBPrintCommands.py
kastiglione: I left a couple comments. Primarily needs a fix for the location of `dataFile`.

commands/FBPrintCommands.py
mattjgalloway: @VTopoliuk actually pointed out that this is required outside of the `elif`, as it's used down on line 476.

kastiglione: @mattjgalloway Ah I see. I didn't know that sandbox apps could write to /tmp. If that works, then sweet!

commands/FBPrintCommands.py
```diff
@@ -424,3 +425,60 @@ def run(self, arguments, options):
...
+          return False
+
+    commandString = ''
+    if dataAsString is not None and len(dataAsString) > 0:
+      commandString += 'echo "{}" | base64 -D -o "{}"; '.format(dataAsString, dataFile)
```
kastiglione: This could be `&&` instead of `;` to prevent curl from trying if the base64 decoding fails.

commands/FBPrintCommands.py
mattjgalloway: Hmm. Actually, it would be best to set to `None` here, and then inside the `elif not runtimeHelpers.isIOSDevice()` set it to the string. That way it's only set if we did actually write it to the file.

commands/FBPrintCommands.py
```diff
@@ -424,3 +425,60 @@ def run(self, arguments, options):
...
+    HTTPData = fb.evaluateObjectExpression('[{} HTTPBody]'.format(request))
+    dataFile = None
+    dataAsString = None
+    if fb.evaluateIntegerExpression('[{} length]'.format(HTTPData)) > 0:
+        dataFile = '/tmp/curl_data_{}'.format(fb.evaluateExpression('(NSTimeInterval)[NSDate timeIntervalSinceReferenceDate]'))
+        if options.embed:
+          if fb.evaluateIntegerExpression('[{} respondsToSelector:@selector(base64EncodedStringWithOptions:)]'.format(HTTPData)):
+            dataAsString = fb.evaluateExpressionValue('(id)[(id){} base64EncodedStringWithOptions:0]'.format(HTTPData)).GetObjectDescription()
+        elif not runtimeHelpers.isIOSDevice():
+          fb.evaluateExpression('(BOOL)[{} writeToFile:@"{}" atomically:NO]'.format(HTTPData, dataFile))
```
mattjgalloway: It would be nice to at least warn if this fails.

VTopoliuk: Oh, you are totally right!

kastiglione:
Looks like you'll need to merge master into this branch now.

kastiglione:
@VTopoliuk shall we merge this? cc @mattjgalloway

mattjgalloway:
Let's do it.

PR #111, opened 2015-09-10
Add print JSON command.

alexch2000:
Add ability to print JSON formatted NSDictionary and NSArray. 
Adding to pasteboard will be added soon
   
```Objective-C
NSDictionary* dictionary = @{@"Array" : @[@1, @"a", @2], @"Value" : @13, @"String" : @"String"};
```
*Example*
**(lldb) pjson dictionary**
```JSON
{
  "Value" : 13,
  "String" : "String",
  "Array" : [
    1,
    "a",
    2
  ]
}
```

### DIFF
--- a/commands/FBPrintCommands.py
+++ b/commands/FBPrintCommands.py
@@ -444,12 +444,8 @@ class FBPrintJSON(fb.FBCommand):
 
   def run(self, arguments, options):
     objectToPrint = arguments[0]
-    pretty = 1 if None is options.plain else 0
+    pretty = 1 if options.plain is None else 0
     jsonData = fb.evaluateObjectExpression('[NSJSONSerialization dataWithJSONObject:{} options:{} error:nil]'.format(objectToPrint, pretty))
     jsonString = fb.evaluateExpressionValue('(NSString*)[[NSString alloc] initWithData:{} encoding:4]'.format(jsonData)).GetObjectDescription()
     
     print jsonString
-
-
-
-
--- a/commands/FBPrintCommands.py
+++ b/commands/FBPrintCommands.py
@@ -32,6 +32,7 @@ def lldbcommands():
     FBPrintApplicationDocumentsPath(),
     FBPrintData(),
     FBPrintTargetActions(),
+    FBPrintJSON(),
   ]
 
 class FBPrintViewHierarchyCommand(fb.FBCommand):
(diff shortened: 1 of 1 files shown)


ilkera:
Thanks Alex for doing this. It will be so helpful!

VTopoliuk:
Hey @kastiglione, actually I would like to have this command without auto-copy to pboard feature. I will not be happy to know my pboard was overridden by any reason. Debugging is iterative process and pboard can contain some important address or string. What about leaving this command as is?

kastiglione:
@VTopoliuk there are other commands that overwrite the pasteboard, which I tend to think of as generally a good thing, but you make a good point. We should either add a global setting, so that users can control this. Alternatively we could add a flag to commands to allow people to enable/disable pasteboard functionality. I think a global setting is better than per-command, what do you think?

VTopoliuk:
@kastiglione I'm totally agree with you. Making global settings will be nice feature, but have per-command support is good feature too. For example I'm not interested on copying result to pboard by default, but sometimes I can override it with flag. So having both features is even better =)
But what about merging this request as is for now? Looks like very nice feature.

commands/FBPrintCommands.py
```diff
@@ -424,3 +425,31 @@ def run(self, arguments, options):
...
+    objectToPrint = arguments[0]
+    pretty = 1 if None is options.plain else 0
+    jsonData = fb.evaluateObjectExpression('[NSJSONSerialization dataWithJSONObject:{} options:{} error:nil]'.format(objectToPrint, pretty))
+    jsonString = fb.evaluateExpressionValue('(NSString*)[[NSString alloc] initWithData:{} encoding:4]'.format(jsonData)).GetObjectDescription()
+    
+    print jsonString
+
+
+
+
```
kastiglione: A few extra blank lines here.

commands/FBPrintCommands.py
```diff
@@ -424,3 +425,31 @@ def run(self, arguments, options):
...
+    return [
+      fb.FBCommandArgument(arg='plain', short='-p', long='--plain', boolean=True, default=False, help='Plain JSON')
+    ]
+
+  def args(self):
+    return [ fb.FBCommandArgument(arg='object', type='id', help='The NSDictionary or NSArray object to print') ]
+
+  def run(self, arguments, options):
+    objectToPrint = arguments[0]
+    pretty = 1 if None is options.plain else 0
```
kastiglione: Could this be

```python
pretty = 1 if options.plain is None else 0
```

kastiglione:
@VTopoliuk Happy to merge this. @alexch2000 would you mind a small update to for the comments I left. They are minor, so if you're busy I'm ok to land as is.

alexch2000:
@kastiglione Sorry for delay. I have changed code.

kastiglione:
That's no delay!

thanks :sparkles:

PR #113, opened 2015-09-17
add two commands pclassmethod and  pinstancemethod

longv2go:
This command show the class`s methods in a pretty way.  

Options:
  --address/-a ; Print the implementation address of the method
  --instance/-i ; Print the instance methods
  --class/-c ; Print the class methods

You can get the runtime address of the method use option '-a' then do something more with that address, 
such as run lldb command "dis -a <addr>" to disassemble the method.

Examples:
```
(lldb) pme -a self
Class Methods:
No instance methods were found.

Instance Methods:
- (void)outAction:(id)arg0  0x000ee680
- (void)innerAction:(id)arg0  0x000ee6f0
- (void)inges:(id)arg0  0x000ee760
- (void)underGes:(id)arg0  0x000ee7d0
......
```


kastiglione:
Thanks for the pull request!

Why not one command, `pmethods` and let the person pass either a class or an instance?

```
(lldb) pmethods self
- (void)setSignInButton:(id)arg0 
- (id)signInFailureText
- (void)setSignInFailureText:(id)arg0 
- (void).cxx_destruct
......
(lldb) pmethods [self class]
+ (id)test:(id)arg0
```

longv2go:
@kastiglione Now It supply only one command 'pmethods', it take two options, if '-a' set, it would print all class methods and instance mehtods, if '-c' set, it would only print class methods, default it only print intance methods.

kastiglione:
:+1: I will take a look tomorrow (it's the end of the night for me now)

One question: What happens if you call `pmethods XYSomeClass`?

longv2go:
@kastiglione The pmethods can take an Instance of some Class or a Class object as its argument, and It 
would print the same thing. It use the options '-a' '-c' to determine print the class methods or instance methods.

```
(lldb) pme self
- (id)signInService
- (id)usernameTextField
- (bool)isValidUsername:(id)arg0 
......

(lldb) pme [self class]
- (id)signInService
- (id)usernameTextField
- (bool)isValidUsername:(id)arg0 
......

(lldb) pme -a self
+ (id)test:(id)arg0 
- (id)signInService
- (id)usernameTextField
- (bool)isValidUsername:(id)arg0 
- (bool)isValidPassword:(id)arg0 
- (id)signInSignal
......
```

kastiglione:
thanks, will try this out today or tomorrow!

PaulTaykalo:
Is there any way to know, what prevents this PR to be merged in ?:)

commands/FBClassDump.py
```diff
@@ -0,0 +1,190 @@
...
+  return [
+    FBPrintMethods()
+  ]
+
+class FBPrintMethods(fb.FBCommand):
+  def name(self):
+    return 'pmethods'
+
+  def description(self):
+    return 'Print the class instance methods.'
```
fbdeveloper: "Print the class and instance methods of a class."

commands/FBClassDump.py
```diff
@@ -0,0 +1,190 @@
...
+
+  def options(self):
+    return [
+      fb.FBCommandArgument(short='-a', long='--all', arg='all', help='If display all methos include class and instance methods', default=False, boolean=True),
+      fb.FBCommandArgument(short='-c', long='--class', arg='clsmethod', help='Print the class methods only', default=False, boolean=True)
+    ]
+
+  def args(self):
+    return [ fb.FBCommandArgument(arg='class or instance', type='id or Class', help='an Object-C Class.') ]
```
fbdeveloper: "Object-C" should be "Objective-C"

commands/FBClassDump.py
```diff
@@ -0,0 +1,190 @@
...
+
+  def run(self, arguments, options):
+    cls = arguments[0]
+    if not isClassObject(cls):
+        cls = runtimeHelpers.object_getClass(cls)
+        if not isClassObject(cls):
+            raise Exception('parameter invalide, not a id or Class')
```
fbdeveloper: "Invalid argument. Please specify an instance or a class."

commands/FBClassDump.py
fbdeveloper: "id or Class" should be "instance or Class."

kastiglione:
@PaulTaykalo I dropped the ball on following up on the review, apologies.

commands/FBClassDump.py
```diff
@@ -0,0 +1,190 @@
...
+class FBPrintMethods(fb.FBCommand):
+  def name(self):
+    return 'pmethods'
+
+  def description(self):
+    return 'Print the class instance methods.'
+
+  def options(self):
+    return [
+      fb.FBCommandArgument(short='-a', long='--all', arg='all', help='If display all methos include class and instance methods', default=False, boolean=True),
```
fbdeveloper: I find the combination of '-a' and '-c' to be a bit counter-intuitive. What about '-i/--instance' for instance and '-c/--class' for class and if neither is specified then it defaults to both. This ensures that there aren't any combinations that don't make sense.

commands/FBClassDump.py
```diff
@@ -0,0 +1,190 @@
...
+    else:
+        printInstanceMethods(cls)
+
+def isClassObject(arg):
+    return runtimeHelpers.class_isMetaClass(runtimeHelpers.object_getClass(arg))
+
+def printInstanceMethods(cls, prefix='-'):
+    ocarray = instanceMethosOfClass(cls)
+    if not ocarray:
+      print "-- have none method or an error occur-- "
```
fbdeveloper: How about: "No instance methods were found."

commands/FBClassDump.py
```diff
@@ -0,0 +1,190 @@
...
+      if method is not None:
+        methods.append(method)
+        print prefix + ' ' + method.prettyPrint()
+
+def printClassMethods(cls):
+    printInstanceMethods(runtimeHelpers.object_getClass(cls), '+')
+
+# I find that a method that has variable parameter can not b.evaluateExpression
+# so I use numberWithLongLong: rather than -[NSString stringWithFormat:]
+def instanceMethosOfClass(klass):
```
fbdeveloper: Typo s/Methos/Methods/

commands/FBClassDump.py
```diff
@@ -0,0 +1,190 @@
...
+      method = createMethodFromOCMethod(i)
+      if method is not None:
+        methods.append(method)
+        print prefix + ' ' + method.prettyPrint()
+
+def printClassMethods(cls):
+    printInstanceMethods(runtimeHelpers.object_getClass(cls), '+')
+
+# I find that a method that has variable parameter can not b.evaluateExpression
+# so I use numberWithLongLong: rather than -[NSString stringWithFormat:]
```
fbdeveloper: This is correct. Can you change the comment to be a bit more assertive? Maybe "Use numberWithLongLong: rather than -[NSString stringWithFormat:] since evaluateExpression doesn't work with variable arguments."

commands/FBClassDump.py
```diff
@@ -0,0 +1,190 @@
...
+  """
+
+  command = string.Template(tmpString).substitute(cls=klass)
+  command = '({' + command + '})'
+  ret = fb.evaluateExpressionValue(command)
+  if not ret.GetError().Success():
+    return None
+
+  ret = ret.GetValue()
+  if int(ret, 16) == 0: # return nil
```
fbdeveloper: This could be removed if you use fb.evaluateIntegerExpressionValue a few lines above.

commands/FBClassDump.py
```diff
@@ -0,0 +1,190 @@
...
+  }
+
+  def __init__(self, name, type_encoding, imp, oc_method):
+    self.name = name
+    self.type = type_encoding
+    self.imp = imp
+    self.oc_method = self.toHex(oc_method)
+
+  def prettyPrint(self):
+    # mast be bigger then 2, 0-idx for self, 1-st for SEL
```
fbdeveloper: s/mast/must

I don't quite understand this comment, especially the "1-st" part.

fbdeveloper:
This is great! Thanks for doing this; I always wanted to do it but never got around to it. I left some comments inline for some changes but other than those it looks great!

longv2go:
Thanks for your comments!!! @fbdeveloper

commands/FBClassDump.py
longv2go: Good idea!

longv2go: I just add a new '-a' option to show the implementation address of the method

commands/FBClassDump.py
longv2go: the fb.evaluateIntegerExpressionValue would round the express with "(int)( expr )"， this can cause error when the expr is "({ ... })"

kastiglione:
Thanks for reviewing @fbdeveloper! By the way, have we met before?

longv2go:
Hi, @fbdeveloper @kastiglione @PaulTaykalo , I change the implementation of pmethods, the CLI not changed. The last version has some issues, it run slowly when the class has many methods and it`s implementation is not so clearly. So I wrote a new method 'evaluate' in fblldbbase.py, it can evaluate a batch of Objective-C expressions and automatic transform the Objective-C object to Python object. I rewrote pmethods base on 'evaluate'. The lines of code go to down, and it more clearly now. Please review this. :)

arigrant:
Ah sorry, that was me, accidentally logged into an old fb dev account for a repo I was looking into.

jamesgpearce:
hahaha

commands/FBClassDump.py
```diff
@@ -0,0 +1,163 @@
...
+    '*': 'char *',
+    '@': 'id',
+    '#': 'Class',
+    ':': 'SEL',
+  }
+
+  def __init__(self, json):
+    self.name = json['name']
+    self.type_encoding = json['type_encoding']
+    self.peremeters_type = json['parameters_type']
```
PaulTaykalo: parameters_type

commands/FBClassDump.py
```diff
@@ -0,0 +1,163 @@
...
+    NSMutableArray *result2 = (id)[NSMutableArray array];
+    
+    for (int i = 0; i < outCount; i++) {
+         
+        NSMutableDictionary *m = (id)[NSMutableDictionary dictionary];
+        
+        SEL name = (SEL)method_getName(methods[i]);
+        [m setObject:(id)NSStringFromSelector(name) forKey:@"name"];
+        
+        char * ecodeing = (char *)method_getTypeEncoding(methods[i]);
```
PaulTaykalo: encoding

commands/FBClassDump.py
```diff
@@ -0,0 +1,163 @@
...
+
+
+  def decode(self, type):
+    ret = type
+    if type in Method.encodeMap:
+      ret = Method.encodeMap[type]
+    return ret
+
+  def toHex(self, addr):
+    return hex(addr)
```
kastiglione: This method looks superfluous.

fblldbbase.py
```diff
@@ -73,3 +74,63 @@ def evaluateExpression(expression, printErrors=True):
...
+
+
+RETURN_MACRO = """
+#define IS_JSON_OBJ(obj)\
+    (obj != nil && ((bool)[NSJSONSerialization isValidJSONObject:obj] ||\
+    (bool)[obj isKindOfClass:[NSString class]] ||\
+    (bool)[obj isKindOfClass:[NSNumber class]]))
+#define RETURN(ret) ({\
+    if (!IS_JSON_OBJ(ret)) {\
+        (void)[NSException raise:@"Invalied RETRUN argument" format:@""];\
```
PaulTaykalo: Invalid RETURN argument

fblldbbase.py
```diff
@@ -73,3 +74,63 @@ def evaluateExpression(expression, printErrors=True):
...
+    NSData *__data = (id)[NSJSONSerialization dataWithJSONObject:__dict options:0 error:NULL];\
+    NSString *__str = (id)[[NSString alloc] initWithData:__data encoding:4];\
+    (char *)[__str UTF8String];})
+#define RETURNCString(ret)\
+    ({NSString *___cstring_ret = [NSString stringWithUTF8String:ret];\
+    RETURN(___cstring_ret);})
+"""
+
+def check_expr(expr):
+  return expr.strip().split('\n')[-1].find('RETURN') != -1
```
kastiglione: This could be more efficient and also simple by comparing the results of `rfind` for `'RETURN'` and `'\n'`.

commands/FBClassDump.py
```diff
@@ -0,0 +1,163 @@
...
+        method = Method(m)
+
+        if showaddr:
+          print prefix + ' ' + method.prettyPrintString() + ' ' + str(method.imp)
+        else:
+          print prefix + ' ' + method.prettyPrintString()
+
+
+def printClassMethods(cls, showaddr=False):
+    printInstanceMethods(runtimeHelpers.object_getClass(cls), showaddr, '+')
```
kastiglione: The above three functions all use an initial 4 space indent, which should be 2 spaces in sticking with the project's default.

fblldbbase.py
PaulTaykalo: In case of objective-C code its's better to split by `;` instead of newlines
Also it's probably will be better to check expression by using regexp on the last expression
Something like
```
/\s*RETURN\w*\s*\(.*)\s*\)\s*;\s*/
```
But searching for `RETURN` itself - is, probably fine

fblldbbase.py
```diff
@@ -73,3 +74,63 @@ def evaluateExpression(expression, printErrors=True):
...
+
+# evaluate a batch of Objective-C expressions, the last expression must contain a RETURN marco
+# and it will automatic transform the Objective-C object to Python object
+# Example:
+#       >>> fblldbbase.evaluate('NSString *str = @"hello world"; RETUTN(@{@"key": str});')
```
PaulTaykalo: RETURN

commands/FBClassDump.py
```diff
@@ -0,0 +1,163 @@
...
+      printInstanceMethods(cls, options.showaddr)
+
+    if not options.clsmethod and not options.insmethod:
+      print 'Class Methods:'
+      printClassMethods(cls, options.showaddr)
+      print '\nInstance Methods:'
+      printInstanceMethods(cls, options.showaddr)
+
+def isClassObject(arg):
+    return runtimeHelpers.class_isMetaClass(runtimeHelpers.object_getClass(arg))
```
PaulTaykalo: There's simple `object_isClass` method in objc_runtime
No need to tweak around with metaclasses

longv2go:
Thanks for reviewing! Sorry for my pool english!

commands/FBClassDump.py
longv2go: It support since iOS 8

kastiglione:
@arigrant I'll be traveling for the next few days. Would you mind merging this when ready?

Thanks @longv2go!

commands/FBClassDump.py
kastiglione: @longv2go Do you plan to address this?

fblldbbase.py
kastiglione: @longv2go This is still in need of fixing.

fblldbobjcruntimehelpers.py
```diff
@@ -32,6 +32,10 @@ def class_getSuperclass(klass):
   value = fb.evaluateExpression(command)
   return value
 
+def class_isMetaClass(klass):
+    command = '(BOOL)class_isMetaClass((Class){})'.format(klass)
+    return fb.evaluateBooleanExpression(command)
```
kastiglione: @longv2go You don't need the `(BOOL)` cast when calling `evaluateBooleanExpression`.

kastiglione:
Thanks @longv2go!

PR #117, opened 2015-10-11
Use base10 value for ivarOffset to be consistent with other values used

idrisr:
in calculating memory address of watchpoint.

When running this line `    watchpoint = lldb.debugger.GetSelectedTarget().WatchAddress(objectAddress + ivarOffset, ivarSize, False, True, error)`

`objectAddress` and `ivarSize` are in base10, while `ivarOffset` is in hex after returning from `fb.evaluateIntegerExpression`.

This fixes that issue and now the memory address of the watchpoint is being set correctly.


kastiglione:
:sparkles: thanks!

PR #119, opened 2015-10-12
Fix bug that was using base16 for int expressions instead of base10

idrisr:
I took a look at all the code that is using `fb.evaluateIntegerExpression` and all of it requires base10 and not base16. This probably wasn't found earlier because most times the argument for the function will be less than 10 and there will be no difference between the base10 and base16 value.

```
id@idmac: ~/chisel
 $ grep -ri IntegerExpression .                                                                                                                                                                 ./commands/FBAccessibilityCommands.py:      accessibilityElementsCount = fb.evaluateIntegerExpression('[%s count]' % accessibilityElements)
./commands/FBFlickerCommands.py:  numViews = fb.evaluateIntegerExpression('[(id)' + subviews + ' count]')
./commands/FBFlickerCommands.py:  numViews = fb.evaluateIntegerExpression('[(id)' + subviews + ' count]')
./commands/FBFlickerCommands.py:  idx = fb.evaluateIntegerExpression('[(id)' + subviews + ' indexOfObject:' + view + ']')
./commands/FBPrintCommands.py:    targetCount = fb.evaluateIntegerExpression('[{targets} count]'.format(targets=targets))
./commands/FBPrintCommands.py:    HTTPHeadersCount = fb.evaluateIntegerExpression('[{} count]'.format(HTTPHeaders))
./commands/FBPrintCommands.py:    if fb.evaluateIntegerExpression('[{} length]'.format(HTTPData)) > 0:
./commands/FBPrintCommands.py:          if fb.evaluateIntegerExpression('[{} respondsToSelector:@selector(base64EncodedStringWithOptions:)]'.format(HTTPData)):
./fblldbbase.py:def evaluateIntegerExpression(expression, printErrors=True):
./fblldbbase.py:  return (int(evaluateIntegerExpression('(BOOL)(' + expression + ')', printErrors)) != 0)
./fblldbviewcontrollerhelpers.py:  numChildViewControllers = fb.evaluateIntegerExpression('(int)[(id)[%s childViewControllers] count]' % (vc))
```

### DIFF
--- a/commands/FBDebugCommands.py
+++ b/commands/FBDebugCommands.py
@@ -32,7 +32,7 @@ class FBWatchInstanceVariableCommand(fb.FBCommand):
     objectAddress = int(fb.evaluateObjectExpression(commandForObject), 0)
 
     ivarOffsetCommand = '(ptrdiff_t)ivar_getOffset((void*)object_getInstanceVariable((id){}, "{}", 0))'.format(objectAddress, ivarName)
-    ivarOffset = fb.evaluateIntegerExpression(ivarOffsetCommand)
+    ivarOffset = int(fb.evaluateExpression(ivarOffsetCommand), 0)
 
     # A multi-statement command allows for variables scoped to the command, not permanent in the session like $variables.
     ivarSizeCommand = ('unsigned int size = 0;'


fblldbbase.py
```diff
@@ -63,7 +63,7 @@ def evaluateIntegerExpression(expression, printErrors=True):
     output = output[2:]
   elif output.startswith('\\'): # Or as \0 (Dec)
     output = output[1:]
-  return int(output, 16)
+  return int(output, 10)
```
kastiglione: Is there a reason not to make this `0` so that it handles decimal and hex?

kastiglione:
Thanks for looking into this!

kastiglione:
There are a bunch of places using `int()` in combination with the `fb.evaluate*` functions, seemingly to avoid this function's hardcoded 16. Would you be interested in switching those to use this fixed `evaluateIntegerExpression`?

fblldbbase.py
idrisr: Good point. I agree it would be better to make this `int(output, 0)` to handle any integer literal.

```
Python 3.4.3 (default, Jul 13 2015, 12:18:23)
[GCC 4.2.1 Compatible Apple LLVM 6.1.0 (clang-602.0.53)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> int('0o13217', 0)
5775
>>> int('0xf3217', 0)
995863
>>> int('0b1010110', 0)
86
>>> int('123', 0)
123
>>>
```

idrisr:
I agree it would be cleaner to replace the `int(fb.evaluate*)` calls with the new `fb.evaluateIntegerExpression` method. I'll put a new PR in soon.

Issue #80, opened 2015-03-18
Issues running in Swift


nickkjordan:
I am getting errors because my expressions are running in Swift despite the [`evaluateExpressionValueInFrameLanguage`](https://github.com/facebook/chisel/blob/abcbb93abe22d98b273c981c71c96ece80286af6/fblldbbase.py#L49) function

Typical error statement:

error: expected ',' separator
(int)((BOOL)((unsigned long)CFGetTypeID((CFTypeRef)(shape)) == (unsigned long)CGImageGetTypeID()))

Is there something I'm missing here?  I'm just calling `visualize <value>`

donshefer:
I'm seeing something similar in my Swift project:

(lldb) caflush
error: <EXPR>:1:22: error: expected ',' separator
(void)[CATransaction flush]

kastiglione:
Thanks for the reports! I will try to take a look at this soon.

yxztj:
Hi, I'm also seeing this error:
(lldb) pviews
error: <EXPR>:1:20: error: expected ',' separator
(id)[(id)[UIWindow keyWindow] recursiveDescription]

kastiglione:
@yxztj I just tried running `pviews` using the "Master-Detail Application" template that comes with Xcode. I could not reproduce the error you've reported. I am using the latest Xcode GM, 6.3.2 GM. If you're able to try the GM and let me know whether the GM fixes this for you, that would be great. Or, using whichever version you have installed currently, could you try creating a new project based on the "Master-Detail Application" template and let me know if `pviews` is broken? Thanks!

kastiglione:
@donshefer I just tried `caflush` using Xcode 6.3.2 GM, and could not reproduce the error you reported. When you update to 6.3.2, if you try `caflush` on your Swift project and it still fails, I would love to know and investigate more. Thanks!

kastiglione:
@nickkjordan I am seeing different errors when I run `visualize`. I'll look into them. I was calling `visualize` on a view. Were you calling it on a image, view, or layer?

nickkjordan:
This is bad, but I can't quite remember at this point.  I *think* it was on a layer though, because I remember doing some animations around then

ignat980:
I am also getting this error. 
```
(lldb) caflush
error: <EXPR>:1:22: error: expected ',' separator
(void)[CATransaction flush]
                     ^
                    ,
```
[...]

PR #121, opened 2015-10-15
Get Swifty - Update everything to use ObjC++ or Swift contexts explicitly

ultramiraculous:
Adding a `po`-like shortcuts to run `expression` in either a Swift (`pswift`) or ObjC++ (`pobjc`) context, rather than use the language of the current frame.  Currently if you're trying do do something like `po [UIViewController _printHierarchy]` and you're not in an ObjC frame, the expression won't parse/run. It's potentially useful if you're trying to re-use commands and you're potentially going to change between languages in the meantime. 

Mainly trying to fix commands like `pvc` that end up on the fast track to Failtown when run from a breakpoint in Swift code.

### DIFF
--- a/commands/FBAccessibilityCommands.py
+++ b/commands/FBAccessibilityCommands.py
@@ -78,9 +78,9 @@ def forceStartAccessibilityServer():
   if not fb.evaluateBooleanExpression('[UIView instancesRespondToSelector:@selector(_accessibilityElementsInContainer:)]'):
     #Starting accessibility server is different for simulator and device
     if fb.evaluateExpressionValue('(id)[[UIDevice currentDevice] model]').GetObjectDescription().lower().find('simulator') >= 0:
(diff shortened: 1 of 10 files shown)


ultramiraculous:
Also probably resolves #80.

tomaskraina:
:+1:

ignat980:
How long does it take to merge a fix? This is a very useful plugin and I'd like for it to work within swift contexts.

arigrant:
This is awesome. Thanks!

ultramiraculous:
(Y)

tomaskraina:
Don't you guys want to make a release (1.4.0?) so we can install/update this using `brew intall chisel`? :) Let me know if I can help with that.

kastiglione:
Thanks @tomaskraina – I'll get right on that.

tomaskraina:
Thank you!

PR #122, opened 2015-10-15
Better Handing of Integer Literals

idrisr:
This is a follow up to https://github.com/facebook/chisel/pull/119. Now `evaluateIntegerExpression()` will accept an integer literal of any base and return the base10 value. Also made code more consistent to call `evaluateIntegerExpression()` where possible instead of `int(evaluateExpression())`

### DIFF
--- a/fblldbbase.py
+++ b/fblldbbase.py
@@ -63,7 +63,7 @@ def evaluateIntegerExpression(expression, printErrors=True):
     output = output[2:]
   elif output.startswith('\\'): # Or as \0 (Dec)
     output = output[1:]
-  return int(output, 16)
+  return int(output, 0)
 
 def evaluateBooleanExpression(expression, printErrors=True):
   return (int(evaluateIntegerExpression('(BOOL)(' + expression + ')', printErrors)) != 0)
--- a/commands/FBAccessibilityCommands.py
+++ b/commands/FBAccessibilityCommands.py
@@ -96,7 +96,7 @@ def printAccessibilityHierarchy(view, indent = 0):
     print indentString + ('{} {}'.format(classDesc, view))
     #We call private method that gives back all visible accessibility children for view
     accessibilityElements = fb.evaluateObjectExpression('[[[UIApplication sharedApplication] keyWindow] _accessibilityElementsInContainer:0 topLevel:%s includeKB:0]' % view)
-    accessibilityElementsCount = int(fb.evaluateExpression('(int)[%s count]' % accessibilityElements))
+    accessibilityElementsCount = fb.evaluateIntegerExpression('(int)[%s count]' % accessibilityElements)
     for index in range(0, accessibilityElementsCount):
       subview = fb.evaluateObjectExpression('[%s objectAtIndex:%i]' % (accessibilityElements, index))
       printAccessibilityHierarchy(subview, indent + 1)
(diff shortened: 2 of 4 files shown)


fblldbbase.py
```diff
@@ -63,7 +63,7 @@ def evaluateIntegerExpression(expression, printErrors=True):
     output = output[2:]
   elif output.startswith('\\'): # Or as \0 (Dec)
     output = output[1:]
-  return int(output, 16)
+  return int(output, 0)
```
erydo: Because the stripped prefix indicates the base of the literal, it seems like it would make more sense to preserve that information rather than asking `int()` to attempt to automatically detect.

For example, if the output were `\x20` (decimal 32) and the prefix were stripped, `int()` would assume `20` was in base 10.

Suggestion:
```py
if output.startswith('\\x'): # Booleans may display as \x01 (Hex)
    output = output[2:]
    base = 16
elif output.startswith('\\'): # Or as \0 (Dec)
    output = output[1:]
    base = 10
else:
    base = 10
  return int(output, base)
```

Three additional comments, from my own unfamiliarity:
1. Finally, the `0` prefix is usually used to indicate octal, not decimal. Is that different here?
2. Because the expression given to LLDB is wrapped in `(int)(…)` above, doesn't LLDB *always* return that value in decimal? It isn't attempting to render a pointer type.
3. To accommodate cases of e.g. a `ptrdiff_t` which may be longer than an int, should the above use `(long long)` rather than `(int)`?

idrisr:
1. I *believe* that LLDB will return the expression cast to `int` in decimal, though I am not certain enough on this to say this is true in all cases. 

2. The `0` prefix does indicate a octal usually, but not for this python command. The second argument is the numeric value of the base you want, so if you wanted octal, the 2nd argument would be `8`.  I'm including the `int()` docstring for Python below.

```
In [1]: int?
Docstring:
int(x=0) -> int or long
int(x, base=10) -> int or long

Convert a number or string to an integer, or return 0 if no arguments
are given.  If x is floating point, the conversion truncates towards zero.
If x is outside the integer range, the function returns a long instead.

If x is not a number or if base is given, then x must be a string or
Unicode object representing an integer literal in the given base.  The
literal can be preceded by '+' or '-' and be surrounded by whitespace.
The base defaults to 10.  Valid bases are 0 and 2-36.  Base 0 means to
interpret the base from the string as an integer literal.
```

3. That's a good point. `ptrdiff_t` can be larger than an `int`. https://www.gnu.org/software/libc/manual/html_node/Important-Data-Types.html

erydo:
@idrisr — Re `0` for octal, I was not referring to the argument to `int`. I was referring to the comment in the line:
```py
elif output.startswith('\\'): # Or as \0 (Dec)
```
Which appears to erroneously interpret LLDB output with a leading `\0` as representing decimal when it likely should represent octal and use `base = 8` instead in the suggestion above.

idrisr:
Yeah that comment is misleading. '0' indicates an octal string literal.

kastiglione:
So is it correct that the remaining changes here are:

- [ ] Preserve base information from lldb output  
- [ ] Fix or remove incorrect "Dec" comment when actually handle octal
- [ ] Cast to `long long`

kastiglione:
Thanks @idrisr and @erydo for quality on this!

PR #126, opened 2015-11-03
Adding --depth option to border command to recursively border subviews

eithanshavit:


kastiglione:
Interesting idea, thanks! We'll get this reviewed soon.

commands/FBDisplayCommands.py
```diff
@@ -40,21 +54,43 @@ def args(self):
...
+      for view, level in viewHelpers.subviewsOfView(args[0]):
+        if level > depth:
+           break
+        if prevLevel != level:
+          color = self.nextColorAfterColor(color)
+          prevLevel = level
+        layer = viewHelpers.convertToLayer(view)
+        setBorder(layer, options.width, color, colorClassName)
+    else:
+      assert depth <= 0, "Recursive bordering is only supported for UIViews"
```
kastiglione: Did you want to have two separate asserts? One on the class, and another on depth?

fblldbviewhelpers.py
```diff
@@ -55,6 +55,24 @@ def convertToLayer(viewOrLayer):
...
+# Yields a tuple of the current view in the tree and its level (view, level)
+def subviewsOfView(view):
+  views = [(view, 0)]
+  yield views[0]
+  while views:
+    (view, level) = views.pop(0)
+    subviews = fb.evaluateExpression('(id)[%s subviews]' % view)
+    subviewsCount = int(fb.evaluateExpression('(int)[(id)%s count]' % subviews))
+    if subviewsCount > 0:
+      for i in range(0, subviewsCount):
```
kastiglione: The `if` isn't necessary, the for `for` loop will do nothing on an empty range.

commands/FBDisplayCommands.py
```diff
@@ -40,21 +54,43 @@ def args(self):
...
+      layer = viewHelpers.convertToLayer(args[0])
+      setBorder(layer, options.width, color, colorClassName)
+
     lldb.debugger.HandleCommand('caflush')
 
+  def nextColorAfterColor(self, color):
+    assert color in self.colors, "{} is not a supported color".format(color)
```
kastiglione: This assert shouldn't be necessary, since it's already checked earlier in the `run` method.

fblldbviewhelpers.py
eithanshavit: Thanks, I adopted this code from one of the other files, so overlooked that.

commands/FBDisplayCommands.py
eithanshavit: I think this method shouldn't rely on being called from `run`, so it needs to protect itself by asserting as well.

commands/FBDisplayCommands.py
eithanshavit: Can you please elaborate a bit, or give an example? Notice that I've changed the code a bit in a new commit, but assert still in place.

kastiglione: This looks like it should be something more like:

```python
assert true, "Recursive bordering is only supported for UIViews"
assert depth <= 0, "Recursive bordering requires a depth greater than 0"
```

eithanshavit: There might be a misunderstanding here. The purpose of this assert is to make sure recursive bordering is not used for layers (or anything that's not a UIView or NSView). However regular bordering is still permitted for layers. So if you're in the code path that takes you to this else block, we should validate that `depth` is not provided to make sure recursive bordering won't happen.

kastiglione: I see. It reads like it's assert bad values on the depth, yet giving an unrelated error message.

kastiglione: I see. It looks like it's assert bad values on the depth, yet giving an unrelated error message.

eithanshavit: Ok, I'll add a comment that should make it clearer.

kastiglione:
Thanks @eithanshavit :green_heart:

kastiglione:
Oops, wrong button.

PR #128, opened 2015-11-18
Fixing pa11y and fa11y commands for iOS8.1+

gkassabli:
Pa11y and fa11y weren't working since iOS8.1 because of Apple API changes. This fix should work both on newer and older versions

### DIFF
--- a/commands/FBAccessibilityCommands.py
+++ b/commands/FBAccessibilityCommands.py
@@ -73,11 +73,14 @@ class FBFindViewByAccessibilityLabelCommand(fb.FBCommand):
     self.foundElement = False
     self.accessibilityGrepHierarchy(rootView, arguments[0])
 
+def isRunningInSimulator():
+  return ((fb.evaluateExpressionValue('(id)[[UIDevice currentDevice] model]').GetObjectDescription().lower().find('simulator') >= 0) or (fb.evaluateExpressionValue('(id)[[UIDevice currentDevice] name]').GetObjectDescription().lower().find('simulator') >= 0))
+
 def forceStartAccessibilityServer():
   #We try to start accessibility server only if we don't have needed method active
   if not fb.evaluateBooleanExpression('[UIView instancesRespondToSelector:@selector(_accessibilityElementsInContainer:)]'):
     #Starting accessibility server is different for simulator and device
-    if (fb.evaluateExpressionValue('(id)[[UIDevice currentDevice] model]').GetObjectDescription().lower().find('simulator') >= 0) | (fb.evaluateExpressionValue('(id)[[UIDevice currentDevice] name]').GetObjectDescription().lower().find('simulator') >= 0):
+    if isRunningInSimulator():
       lldb.debugger.HandleCommand('eobjc (void)[[UIApplication sharedApplication] accessibilityActivate]')
     else:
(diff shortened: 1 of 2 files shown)


commands/FBAccessibilityCommands.py
```diff
@@ -77,7 +77,7 @@ def forceStartAccessibilityServer():
   #We try to start accessibility server only if we don't have needed method active
   if not fb.evaluateBooleanExpression('[UIView instancesRespondToSelector:@selector(_accessibilityElementsInContainer:)]'):
     #Starting accessibility server is different for simulator and device
-    if fb.evaluateExpressionValue('(id)[[UIDevice currentDevice] model]').GetObjectDescription().lower().find('simulator') >= 0:
+    if fb.evaluateExpressionValue('(id)[[UIDevice currentDevice] name]').GetObjectDescription().lower().find('simulator') >= 0:
```
kastiglione: Is this a backwards compatible change, will it still work < iOS 8.1?

commands/FBAccessibilityCommands.py
```diff
@@ -86,6 +86,15 @@ def accessibilityLabel(view):
   #using Apple private API to get real value of accessibility string for element.
   return fb.evaluateExpressionValue('(id)[%s accessibilityAttributeValue:%i]' % (view, ACCESSIBILITY_LABEL_KEY), False)
 
+def accessibilityElements(view):
+  a11yElements = fb.evaluateExpression('(id)[%s accessibilityElements]' % view, False)
```
kastiglione: This method is only publicly available in iOS 8 and later. Do you know if it's available on iOS 7.x? Ideally we should still support iOS 7.x, or at least should handle it gracefully.

kastiglione:
Thanks @gkassabli! I have a couple questions.

gkassabli:
Thanks for good point, @kastiglione. Addressed your comments!

commands/FBAccessibilityCommands.py
```diff
@@ -77,7 +77,7 @@ def forceStartAccessibilityServer():
   #We try to start accessibility server only if we don't have needed method active
   if not fb.evaluateBooleanExpression('[UIView instancesRespondToSelector:@selector(_accessibilityElementsInContainer:)]'):
     #Starting accessibility server is different for simulator and device
-    if fb.evaluateExpressionValue('(id)[[UIDevice currentDevice] model]').GetObjectDescription().lower().find('simulator') >= 0:
+    if (fb.evaluateExpressionValue('(id)[[UIDevice currentDevice] model]').GetObjectDescription().lower().find('simulator') >= 0) | (fb.evaluateExpressionValue('(id)[[UIDevice currentDevice] name]').GetObjectDescription().lower().find('simulator') >= 0):
```
kastiglione: 1. This should use `or` instead of binary `|`.
2. Maybe extract into a helper method, `isSimulator`?
3. Is there a reason this isn't using `evaluateObjectExpression`?

kastiglione:
Thanks @gkassabli. One more comment, but otherwise looks good!

commands/FBAccessibilityCommands.py
gkassabli: Will address first 2 comments
3. EvaluateObjectExpression will give us string address, while we need string python representation. expressionValue().GetObjectDescription() seems best way for that

kastiglione:
Thank you :gem:

PR #132, opened 2015-12-30
add two new Why commands

Ganem:
Adding 2 new commands: wnvisible and wninteractible

These commands get one argument - a view - and tries to determine why the given view is not visible or not interactible.
Take a look at the comments in the code to see the different scenarions.

cc @eithanshavit - asked to be included
cc @kastiglione - I guess you would also like to have a look

### DIFF
--- a/commands/FBWhyCommands.py
+++ b/commands/FBWhyCommands.py
@@ -9,7 +9,7 @@ import fblldbobjecthelpers as objcHelpers
 def lldbcommands():
   return [ 
     FBPrintWhyViewNotVisible(),
-    FBPrintWhyViewNotInteractible(),
+    FBPrintWhyViewNotInteractable(),
   ]
 
 class FBPrintWhyViewNotVisible(fb.FBCommand):
@@ -137,15 +137,15 @@ class FBPrintWhyViewNotVisible(fb.FBCommand):
       return None
 
 
-class FBPrintWhyViewNotInteractible(fb.FBCommand):
+class FBPrintWhyViewNotInteractable(fb.FBCommand):
   def name(self):
-    return 'wninteractible'
+    return 'wninteractable'
 
   def description(self):
-    return 'Print the reasons the given view is not interactible.'
+    return 'Print the reasons the given view is not interactable.'
 
   def args(self):
-    return [ fb.FBCommandArgument(arg='aView', type='UIView/NSView *', help='The view to check why it isn\'t interactible.') ]
+    return [ fb.FBCommandArgument(arg='aView', type='UIView/NSView *', help='The view to check why it isn\'t interactable.') ]
 
   def run(self, arguments, options):
     view = arguments[0]
--- a/commands/FBWhyCommands.py
+++ b/commands/FBWhyCommands.py
@@ -1,0 +1,179 @@
+#!/usr/bin/python
+# Example file with custom commands, located at /magical/commands/example.py
+
+import lldb
+import fblldbbase as fb
+import fblldbviewhelpers as viewHelpers
+import fblldbobjecthelpers as objcHelpers
+
+def lldbcommands():
+  return [ 
+    FBPrintWhyViewNotVisible(),
+    FBPrintWhyViewNotInteractible(),
+  ]
+
+class FBPrintWhyViewNotVisible(fb.FBCommand):
+  def name(self):
+    return 'wnvisible'
+
+  def description(self):
+    return 'Print the reasons the given view is not visible.'
+
+  def args(self):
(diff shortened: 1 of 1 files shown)


arigrant:
I wonder if we should make the command `whynot`:

`whynot visible aView`
`whynot interactable aView`

this could also be abbreviated to `wn` for easier use. However we don't have really any of these meta commands yet...

@kastiglione, thoughts?

commands/FBWhyCommands.py
```diff
@@ -0,0 +1,179 @@
...
+
+import lldb
+import fblldbbase as fb
+import fblldbviewhelpers as viewHelpers
+import fblldbobjecthelpers as objcHelpers
+
+def lldbcommands():
+  return [ 
+    FBPrintWhyViewNotVisible(),
+    FBPrintWhyViewNotInteractible(),
```
arigrant: s/interactible/interactable/g

Ganem:
Does that mean the user will need to type 'wn v aView'? Right now you can use 'wnv' and 'wni' which is pretty straightforward.

kastiglione:
> Does that mean the user will need to type `wn v aView`?

We could have single word commands as additional aliases. In other words, `whynot <predicate> args…` could be the official interface, with `wnv` and `wni` being aliases to the full commands. Encouraging users to create their own `~/.lldbinit` aliases is another approach.

The question here might be: where do we want to put the "complexity"? Do we give each of these a top level command which results in chisel having a larger and thus less approachable suite of top level commands, or do we add a meta `whynot` command to reduce the top level complexity at the cost of making the `whynot`s more complicated.

(I think I prefer moving complexity into `whynot`s).

commands/FBWhyCommands.py
```diff
@@ -0,0 +1,179 @@
...
+    width = float(fb.evaluateExpression(size + '.width'))
+    height = float(fb.evaluateExpression(size + '.height'))
+    if width == 0.0 or height == 0.0:
+      reasons += 'The width or height is 0. Did you forget to set the frame?\n'
+
+    # view is out of bounds of its superview
+    origin = '((CGRect)[(id)' + view + ' frame]).origin'
+    xValue = float(fb.evaluateExpression(origin + '.x'))
+    yValue = float(fb.evaluateExpression(origin + '.y'))
+    if xValue < 0.0 or yValue < 0.0:
```
kastiglione: This will produce false positives. Negative origin isn't necessarily a cause of views not showing.

Could this use `CGRectIntersectsRect` to see if a view's frame shares space with its superview?

commands/FBWhyCommands.py
```diff
@@ -0,0 +1,179 @@
...
+      reasons += 'The x or y values are smaller than 0\n'
+
+    # view is hidden
+    isHidden = fb.evaluateBooleanExpression('[' + view + ' isHidden]')
+    if isHidden:
+      reasons += 'View\'s hidden property is YES\n'
+
+    # alpha is 0
+    alpha = float(fb.evaluateExpression('(CGFloat)[' + view + ' alpha]'))
+    if alpha == 0.0:
```
kastiglione: Maybe instead of `0`, a small number in case this has a garbage or unintentional value that is not exactly zero but still results in an alpha that makes the view essentially invisible. Also what happens when the alpha is less than zero, is that a case this should check?

commands/FBWhyCommands.py
```diff
@@ -0,0 +1,179 @@
...
+    while obj:
+      if self.isViewController(obj):
+        return obj
+      else:
+        obj = self.nextResponder(obj)
+
+    return None
+
+  def isViewController(self, obj):
+    return objcHelpers.isKindOfClass('(' + obj + ')', 'UIViewController')
```
kastiglione: Can the wrapping parens be moved into `isKindOfClass`?

commands/FBWhyCommands.py
```diff
@@ -0,0 +1,179 @@
...
+
+  def nextResponder(self, obj):
+    aNextResponder = fb.evaluateObjectExpression('[((id){}) nextResponder]'.format(obj))
+    try:
+      if int(aNextResponder, 0):
```
kastiglione: When does `int` throw?

kastiglione:
@Ganem What do you think about checking the transform on the view to make sure its not a cause for an invisible view?

kastiglione:
@Ganem ping

kastiglione:
@Ganem ghost ping 😛

kastiglione:
@Ganem feel free to reopen at any time, but I'm presuming you've moved on from this.

PR #133, opened 2016-01-03
Add two commands pproperties, pblock

longv2go:
pproperties can print the properties of an instance of Class.

```
(lldb) ppro self
@property (nonatomic, assign, readonly) NSString * helo;
@property (nonatomic, strong) NSNumber * number;
```
pblock can print the implementation address and signature of a block.
```
(lldb) pbl block
Imp: 0x101180980    Signature: void ^(NSString *);
```

### DIFF
--- a/commands/FBClassDump.py
+++ b/commands/FBClassDump.py
@@ -30,14 +30,7 @@ class FBPrintMethods(fb.FBCommand):
     return [ fb.FBCommandArgument(arg='class or instance', type='instance or Class', help='an Objective-C Class.') ]
 
   def run(self, arguments, options):
-    cls = arguments[0]
-    if options.clsname:
-      cls = "(Class)NSClassFromString(@\"{}\")".format(cls)
-    else:
-      if not isClassObject(cls):
-        cls = runtimeHelpers.object_getClass(cls)
-        if not isClassObject(cls):
-            raise Exception('Invalid argument. Please specify an instance or a Class.')
+    cls = getClassFromArgument(arguments[0], options.clsname)
 
     if options.clsmethod:
       print 'Class Methods:'
@@ -64,7 +57,6 @@ class FBPrintProperties(fb.FBCommand):
 
   def options(self):
     return [
-      fb.FBCommandArgument(short='-v', long='--value', arg='showvalue', help='Print the value of a property, not support for now', default=False, boolean=True),
       fb.FBCommandArgument(short='-n', long='--name', arg='clsname', help='Take the argument as class name', default=False, boolean=True)
     ]
 
@@ -72,14 +64,7 @@ class FBPrintProperties(fb.FBCommand):
     return [ fb.FBCommandArgument(arg='class or instance', type='id or Class', help='an Objective-C Class.') ]
 
   def run(self, arguments, options):
-    cls = arguments[0]
-    if options.clsname:
-      cls = "(Class)NSClassFromString(@\"{}\")".format(cls)
-    else:
-      if not isClassObject(cls):
-        cls = runtimeHelpers.object_getClass(cls)
-        if not isClassObject(cls):
-            raise Exception('Invalid argument. Please specify an instance or a Class.')
+    cls = getClassFromArgument(arguments[0], options.clsname)
 
     printProperties(cls)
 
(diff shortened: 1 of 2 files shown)


kastiglione:
Thank you. I'm hoping to review this before the end of the week.

kastiglione:
@longv2go What do you think of looking up the symbol that corresponds to the block's `invoke` function, using something like `image lookup --address` or via the equivalent lldb api? Another thought to put out there, what do you think of print the objects that the block captures?

longv2go:
The pblock just show the invoke function`s address, and I do not know which lldb api can do that. I can go deep with the block struct to find out if I can print the objects that the block captures.

longv2go:
@kastiglione The pblock just show the invoke function`s address, and I do not know which lldb api can do that. I can go deep with the block struct to find out if I can print the objects that the block captures.

kastiglione:
I'm also not sure which part of the lldb api we're looking for, but worst case scenario we could call `HandleCommand` and pass it `image lookup --address` with the address of the `invoke` function pointer.

The two suggestions were just for discussion, they actually came up today when I was showing chisel to someone. They asked about inspecting blocks and I remembered this pull request (sorry). If adding symbol lookup isn't much work, which I hope is the case, then I think it's a no brainer to add. However I suspect getting the captured objects is more work and I also don't think it's as valuable as the symbol lookup. Those are my thoughts.

thanks!

commands/FBClassDump.py
```diff
@@ -145,20 +282,79 @@ def prettyPrintString(self):
...
     return hex(addr)
 
   def __str__(self):
     return "<Method:" + self.oc_method + "> " + self.name + " --- " + self.type + " --- " + self.imp
+
+def getProperties(klass):
+  tmpString = """
+      NSMutableArray *result = (id)[NSMutableArray array];
+      unsigned int count;
+      objc_property_t *props = (objc_property_t *)class_copyPropertyList((Class)$cls, &count);
```
kastiglione: `props` does not ever get `free`d.

commands/FBClassDump.py
```diff
@@ -145,20 +282,79 @@ def prettyPrintString(self):
...
+          
+          char *name = (char *)property_getName(props[i]);
+          [dict setObject:(id)[NSString stringWithUTF8String:name] forKey:@"name"];
+          
+          char *attrstr = (char *)property_getAttributes(props[i]);
+          [dict setObject:(id)[NSString stringWithUTF8String:attrstr] forKey:@"attributes_string"];
+          
+          NSMutableDictionary *attrsDict = (id)[NSMutableDictionary dictionary];
+          unsigned int pcount;
+          objc_property_attribute_t *attrs = (objc_property_attribute_t *)property_copyAttributeList(props[i], &pcount);
```
kastiglione: `attrs` does not ever get `free`d.

longv2go:
@kastiglione   I remove the "free()" because it would cause evaluateExpressionValue to raise exception some time. I do not know the reason. I guess it`s a lldb bug.

longv2go:
@kastiglione The block struct is dynamically generated by the compiler. The captured objects and other objects the block referenced are all attached to that struct, such as 
```
struct __block_literal_2 {
    void *isa;
    int flags;
    int reserved;
    void (*invoke)(struct __block_literal_2 *);
    struct __block_descriptor_2 *descriptor;
    const int x;
};
```
The ```const int x;``` is an object referenced by block. So every block is different. I did not find a way to list all captured objects at the runtime.

kastiglione:
Thanks @longv2go. I'll give this another review today or tomorrow.

kastiglione:
> I remove the `free()` because it would cause `evaluateExpressionValue` to raise exception some time on real device.

@longv2go Do you have any more details on this? I guess it doesn't matter if a bit of memory leaks in debugging, but if it is an lldb bug, we might want to report it or track it if it's known.

longv2go:
@kastiglione When I add the ```free()```, I get the error message:

```
error: Execution was interrupted, reason: EXC_BAD_ACCESS (code=257, address=0x188ae089d).
error: Execution was interrupted, reason: EXC_BAD_ACCESS (code=257, address=0x188ae089d).
  File "/Users/didi/Workspace/github/chisel/fblldb.py", line 79, in runCommand
    command.run(args, options)
  File "/Users/didi/Workspace/github/chisel/commands/FBClassDump.py", line 84, in run
    printProperties(cls)
  File "/Users/didi/Workspace/github/chisel/commands/FBClassDump.py", line 195, in printProperties
    props = getProperties(cls)
  File "/Users/didi/Workspace/github/chisel/commands/FBClassDump.py", line 336, in getProperties
    return [Property(m) for m in propsJson]
TypeError: 'NoneType' object is not iterable
```

and the ```fb.evaluate(command)``` get a ```None```.

kastiglione:
Thanks. Would you mind putting that into a gist, and adding a comment with a link where the `free` go? If you're busy don't worry, we can refer back to this pull request.

longv2go:
@kastiglione I create a new repo. I find a way to print all object of a block of using fishhook. Because the evaluateExpression function can not define a c function, I could`t implement chisel command to print those objects. Please refer https://github.com/longv2go/BlockInspectorTest/tree/master#blockinspector

kastiglione:
@longv2go Does your new repo supersede this pull request? Should I still consider merging this, or should people instead use your BlockInspector?

longv2go:
@kastiglione the repo BlockInspector is written in Objective-C, It can not implement as a chisel command.

kastiglione:
@longv2go Understood. Would you like to continue with this pull request? Or do you plan to use BlockInspector instead?

longv2go:
@kastiglione continue this pull request.

kastiglione:
Thanks again!

PR #136, opened 2016-01-16
Add C function to enumerate all live memory allocations

kastiglione:
I am starting to explore debugging commands for querying and aggregating the graph of Objective-C objects. The first step to building the graph is to scan the heap looking for Objective-C objects. This pull request adds `CHLEnumerateAllocationsWithBlock` to do just that. In a subsequent pull request, this function will be used to identify memory that holds Objective-C objects.

I tried to add some sorely needed documentation about `malloc_get_all_zones`, since there is no man page for it, and the header docs in `malloc.h` leave much to be desired. I first became aware of this function when using @ryanolsonk's awesome [FLEX](https://github.com/Flipboard/FLEX/).

### DIFF
--- a/Chisel/CHLFunctions.h
+++ b/Chisel/CHLFunctions.h
@@ -1,0 +1,15 @@
+//
+//  CHLFunctions.h
+//  Chisel
+//
+//  Copyright © 2016 Facebook. All rights reserved.
+//
+
+#import <Foundation/Foundation.h>
+
+#import <malloc/malloc.h>
+
+/**
+ * Enumerates all live memory allocations by calling `block` for each malloc range (represented as a `vm_range_t`).
+ */
+void CHLEnumerateAllocationsWithBlock(void (^block)(vm_range_t range));
--- a/Chisel/CHLFunctions.m
+++ b/Chisel/CHLFunctions.m
@@ -1,0 +1,65 @@
+//
+//  CHLFunctions.m
+//  Chisel
+//
+//  Copyright © 2016 Facebook. All rights reserved.
+//
+
+#import "CHLFunctions.h"
+
+// See vm_range_recorder_t
+static void enumerateRanges(task_t, void *, unsigned, vm_range_t *, unsigned);
+
+void CHLEnumerateAllocationsWithBlock(void (^block)(vm_range_t))
+{
+  /*
+   * === Overview ===
+   *
+   * To enumerate the set of live memory allocations ("ranges"), the pseudocode might look like:
+   *
+   *     for zone in all_zones()
+   *       for range in zone->ranges
+   *         callback(range)
+   *
+   * Unfortunately, the memory allocations within a zone are not exposed, so the above inner loop is not possible.
+   * Instead malloc zones provide a function that will loop over the allocations and call the given callback function.
+   */
+
+  /*
+   * === malloc API Notes ===
+   *
+   * Both `malloc_get_all_zones` and `vm_range_recorder_t` callback functions take a `task_t` and a `memory_reader_t`.
+   * However it turns out both of these can be NULL. This information isn't documented in malloc.h but can be seen in
+   * the libmalloc source.
+   *
+   * `memory_reader_t`:
+   * When NULL, the default memory reader is used which handles in process memory, and does not use the given `task_t`.
+   * `task_t`:
+   * Only required if used in a custom `memory_reader_t`, or used in a `vm_range_recorder_t` callback.
+   */
+
+  // Get a reference to the C array of malloc zones.
+  vm_address_t *zones;
+  unsigned int zoneCount;
+  malloc_get_all_zones(TASK_NULL, NULL, &zones, &zoneCount);
+
+  // Outer loop over the zones.
+  for (unsigned int i = 0; i < zoneCount; i++) {
+    // While `zones` is typed as `vm_address_t *` for type checking purposes, these addresses are just pointers to
+    // `malloc_zone_t` structs. Thus `vm_address_t` and `malloc_zone_t *` are used interchangably here.
+    malloc_zone_t *zone = (malloc_zone_t *)zones[i];
+
+    // Inner loop over the zone's memory allocations.
+    zone->introspect->enumerator(TASK_NULL, block, MALLOC_PTR_IN_USE_RANGE_TYPE, (vm_address_t)zone, NULL, enumerateRanges);
+  }
+}
+
+// This function is called with an array of memory allocations, which are represented as `vm_range_t` structs.
+static void enumerateRanges(__unused task_t task, void *context, __unused unsigned type, vm_range_t *ranges, unsigned rangeCount)
+{
+  void (^block)(vm_range_t) = context;
+
+  for (unsigned int i = 0; i < rangeCount; i++) {
+    block(ranges[i]);
+  }
+}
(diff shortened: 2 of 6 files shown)


kastiglione:
@ryanolsonk @rnystrom @modocache @jspahrsummers @arigrant @coneko tagging you in case you'd like to give feedback here on this pull request, or later when this gets more interesting.

Chisel/CHLFunctions.m
```diff
@@ -0,0 +1,65 @@
...
+   *
+   * Unfortunately, the memory allocations within a zone are not exposed, so the above inner loop is not possible.
+   * Instead malloc zones provide a function that will loop over the allocations and call the given callback function.
+   */
+
+  /*
+   * === malloc API Notes ===
+   *
+   * Both `malloc_get_all_zones` and `vm_range_recorder_t` callback functions take a `task_t` and a `memory_reader_t`.
+   * However it turns out both of these can be NULL. This information isn't documented in malloc.h but can be seen in
```
ryanolsonk: Interesting! Probably makes sense to update FLEX to pass `NULL`s, yea?

Chisel/CHLFunctions.m
```diff
@@ -0,0 +1,65 @@
...
+  malloc_get_all_zones(TASK_NULL, NULL, &zones, &zoneCount);
+
+  // Outer loop over the zones.
+  for (unsigned int i = 0; i < zoneCount; i++) {
+    // While `zones` is typed as `vm_address_t *` for type checking purposes, these addresses are just pointers to
+    // `malloc_zone_t` structs. Thus `vm_address_t` and `malloc_zone_t *` are used interchangably here.
+    malloc_zone_t *zone = (malloc_zone_t *)zones[i];
+
+    // Inner loop over the zone's memory allocations.
+    zone->introspect->enumerator(TASK_NULL, block, MALLOC_PTR_IN_USE_RANGE_TYPE, (vm_address_t)zone, NULL, enumerateRanges);
```
ryanolsonk: Is it possible that either `introspect` or `enumerator` could be `NULL`?

Chisel/CHLFunctions.m
```diff
@@ -0,0 +1,65 @@
...
+  }
+}
+
+// This function is called with an array of memory allocations, which are represented as `vm_range_t` structs.
+static void enumerateRanges(__unused task_t task, void *context, __unused unsigned type, vm_range_t *ranges, unsigned rangeCount)
+{
+  void (^block)(vm_range_t) = context;
```
ryanolsonk: I love how you can just pass a block through the context param. Makes things so clean.

ryanolsonk:
:shipit:

Chisel/CHLFunctions.m
kastiglione: This file compiled with ARC turned off, otherwise there'd be bridge casting on the `context`.

Chisel/CHLFunctions.m
kastiglione: Theoretically yes, in practice it's currently always present. I wanted to try to make this code easy to digest, so I left out the NULL checks here and also don't check the return value of `malloc_get_all_zones`. In part I feel semi justified because this is intended for production use, just debugging. Thoughts?

Chisel/CHLFunctions.m
kastiglione: :+1:

Chisel/CHLFunctions.m
ryanolsonk: Sure, I think I basically cargo culted the `NULL` checks from the few existing places I found using this API.

Chisel/CHLFunctions.m
```diff
@@ -0,0 +1,65 @@
...
+   *
+   * Both `malloc_get_all_zones` and `vm_range_recorder_t` callback functions take a `task_t` and a `memory_reader_t`.
+   * However it turns out both of these can be NULL. This information isn't documented in malloc.h but can be seen in
+   * the libmalloc source.
+   *
+   * `memory_reader_t`:
+   * When NULL, the default memory reader is used which handles in process memory, and does not use the given `task_t`.
+   * `task_t`:
+   * Only required if used in a custom `memory_reader_t`, or used in a `vm_range_recorder_t` callback.
+   */
```
modocache: nit: Seems like these comments should be placed as a docblock above the function; that way you could option-click to see it in Xcode. I'd also swap out the `=== headers ===` in favor of something more Doxygen/Xcode-esque.

ryanolsonk: I think we probably do want to keep these comments internal since they're about the malloc API that's being wrapped. This function provides a nice abstraction so that users don't have to worry about these details.

Chisel/CHLFunctions.m
kastiglione: Cool, cargo cult is exactly what I wanted to avoid.

ryanolsonk: Haha, well it's not so unreasonable to check there. Infer might yell about it.

Chisel/CHLFunctions.m
kastiglione: ^ this.

The info should be in `malloc.h`, but since they're not, I documented the internal implementation.

kastiglione: I just tried re-declaring `malloc_get_all_zones` inside this file, and giving that declaration a docblock to see if this local declaration would override the one in `malloc.h`, and have Xcode show this local docblock on option-click, but it didn't work. I could avoid importing `malloc.h` and create my own header that declares what this file uses, with its own documentation, but I don't think that's really a better approach.

Chisel/CHLFunctions.h
```diff
@@ -0,0 +1,15 @@
...
+//
+
+#import <Foundation/Foundation.h>
+
+#import <malloc/malloc.h>
+
+/**
+ * Enumerates all live memory allocations by calling `block` for each malloc range (represented as a `vm_range_t`).
+ */
+void CHLEnumerateAllocationsWithBlock(void (^block)(vm_range_t range));
```
kastiglione: @modocache What do I add here to indicate the block can't be null? Should I add an assert internally too?

Chisel/CHLFunctions.m
modocache: I understand your "it's describing an external API" philosophical argument. My argument is far more practical: move the comments into a docblock and chisel developers can option-click to see them anytime. If you leave them where they are, they can't.

Some context: especially for relatively small projects, I prefer to make it easy for people to read documentation, even if it makes it harder to maintain that documentation. For example, rather than `see the documentation for -[MyClass myMethod] for details`, I prefer to provide a full explanation for each docblock. Similarly, here I think developers could find this information more easily if it were in a docblock, even if notes on `malloc` don't *really* belong in a docblock for a different function.

That's my personal preference, though--hence the "nit" on my original comment. :wink: 

> I just tried re-declaring `malloc_get_all_zones` inside this file, and giving that declaration a docblock. 

This is definitely overkill in my book.

Chisel/CHLFunctions.h
modocache: If you mean nullability identifiers, you could:

1. Wrap the header in `NS_ASSUME_NONNULL`.
2. Add `nullable` or `_Nullable` (or `__nullable` if you're supporting Xcode 6.4 and prior) to all nullable parameters and return values.

You should still check for null in the implementation. Without certain warnings enabled, it's very easy for users to accidentally pass a nullable pointer to a parameter marked as nonnull; see: http://modocache.io/clang-warn-nullable-to-nonnull-conversion. Even with those warnings enabled, it's still pretty easy, although newer versions of the Clang static analyzers may catch more of these cases.

Chisel/CHLFunctions.m
kastiglione: I'll think about this some more. I'm down with "make it easy for people to read documentation" but the documentation in these comments does not belong in this file's header.

kastiglione:
cc @Gricha

PR #137, opened 2016-02-15
Adding text input command

tecknut:
# Additions 

* Adds an `input` command which sets text on text views or text fields which were first responder when execution paused.
* Adds `setxt` command which sets text on a view by accessibility ID.

### Example 

```
(lldb) input 'Some text to set.'
(lldb) setxt some.accessibility.id 'Some text to set.'
```

### DIFF
--- a/commands/FBTextInputCommands.py
+++ b/commands/FBTextInputCommands.py
@@ -1,0 +1,71 @@
+#!/usr/bin/python
+
+# Copyright (c) 2016, Facebook, Inc.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree. An additional grant
+# of patent rights can be found in the PATENTS file in the same directory.
+
+import os
+
+import lldb
+import fblldbbase as fb
+import fblldbviewhelpers as viewHelpers
+
+ACCESSIBILITY_ID = 0
+REPLACEMENT_TEXT = 1
+
+
+def lldbcommands():
+  return [
+    FBInputTextCommand(),
+  ]
+
+
+class FBInputTextCommand(fb.FBCommand):
+  def name(self):
+    return 'input'
+
+  def description(self):
+    return 'Input text into text field or text view by accessibility id.'
+
+  def args(self):
+    return [
+      fb.FBCommandArgument(arg='accessibilityId', type='string', help='The accessibility ID of the input view.'),
+      fb.FBCommandArgument(arg='replacementText', type='string', help='The text to set.')
+    ]
+
+  def run(self, arguments, options):
+    rootView = fb.evaluateObjectExpression('[[UIApplication sharedApplication] keyWindow]')
+    self.findView(rootView, arguments[ACCESSIBILITY_ID], arguments[REPLACEMENT_TEXT])
+
+  def findView(self, view, searchIdentifier, replacementText):
+    views = self.subviewsOfView(view)
+    for index in range(0, self.viewsCount(views)):
+        subview = self.subviewAtIndex(views, index)
+        self.findView(subview, searchIdentifier, replacementText)
+    else:
+      identifier = self.accessibilityIdentifier(view)
+      if self.isEqualToString(identifier, searchIdentifier):
+        self.setTextInView(view, replacementText)
+
+# Some helpers
+  def subviewsOfView(self, view):
+    return fb.evaluateObjectExpression('[%s subviews]' % view)
+
+  def subviewAtIndex(self, views, index):
+    return fb.evaluateObjectExpression('[%s objectAtIndex:%i]' % (views, index))
+
+  def viewsCount(self, views):
+    return int(fb.evaluateExpression('(int)[%s count]' % views))
+
+  def accessibilityIdentifier(self, view):
+    return fb.evaluateObjectExpression('[%s accessibilityIdentifier]' % view)
+
+  def isEqualToString(self, identifier, needle):
+    return fb.evaluateBooleanExpression('[%s isEqualToString:@"%s"]' % (identifier, needle))
+
+  def setTextInView(self, view, text):
+    fb.evaluateObjectExpression('[%s setText:@"%s"]' % (view, text))
+    viewHelpers.flushCoreAnimationTransaction()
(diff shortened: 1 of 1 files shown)


commands/FBTextInputCommands.py
```diff
@@ -0,0 +1,111 @@
...
+    self.findFirstResponder(rootView(), arguments[INPUT_TEXT])
+
+  def findFirstResponder(self, view, replacementText):
+    views = subviewsOfView(view)
+    for index in range(0, viewsCount(views)):
+        subview = subviewAtIndex(views, index)
+        self.findFirstResponder(subview, replacementText)
+    else:
+      if isFirstResponder(view):
+        setTextInView(view, replacementText)
```
kastiglione: My python is clearly lacking, how does this `else:` section work? Does it only run if the `for` loop has nothing to loop over (in this case view count is zero)?

commands/FBTextInputCommands.py
```diff
@@ -0,0 +1,111 @@
...
+
+
+# Some helpers
+def rootView():
+  return fb.evaluateObjectExpression('[[UIApplication sharedApplication] keyWindow]')
+
+def subviewsOfView(view):
+  if fb.evaluateBooleanExpression('[%s respondsToSelector:@selector(subviews)]' % view):
```
kastiglione: Why do all of these check the selector first?

commands/FBTextInputCommands.py
```diff
@@ -0,0 +1,111 @@
...
+def lldbcommands():
+  return [
+    FBInputTexByAccessibilityIdCommand(),
+    FBInputTexToFirstResponderCommand(),
+  ]
+
+
+class FBInputTexByAccessibilityIdCommand(fb.FBCommand):
+  def name(self):
+    return 'setxt'
```
kastiglione: What is the mnemonics behind this name? What does `xt` mean?

commands/FBTextInputCommands.py
```diff
@@ -0,0 +1,111 @@
...
+        self.findView(subview, searchIdentifier, replacementText)
+    else:
+      identifier = accessibilityIdentifier(view)
+      if isEqualToString(identifier, searchIdentifier):
+        setTextInView(view, replacementText)
+
+
+class FBInputTexToFirstResponderCommand(fb.FBCommand):
+  def name(self):
+    return 'input'
```
kastiglione: Can has verb? Ideally I would call this `settext` but that is too close to `setxt` (though see my comment on that name too).

kastiglione:
Thanks @tecknut for the pull request!

commands/FBTextInputCommands.py
tecknut: Yes that's exactly it. My python isn't that great either, only found out about this while writing this code.

commands/FBTextInputCommands.py
tecknut: There was one instance where an unrecognised selector erred out, so I decided to be defensive and guard them. 

I'm happy to change this if there's a strong opinion. But there is at least one which caused a stuff to fail, can't remember off hand though.

commands/FBTextInputCommands.py
tecknut: I chose input because this acts like someone inputing text into a text field/text view which has the cursor in it. 

I could change this to `inputtext` and the previous command to `settext`. 

I went with terse commands based on whats currently there in lldb and chisel. Again, no hard opinion on what these should be called.

commands/FBTextInputCommands.py
tecknut: It's just short for text. As per the comment bellow, I based the name on other examples of short hand commands.

commands/FBTextInputCommands.py
kastiglione: This won't handle any first responder that has subviews. Is there a reason for that? I'm not aware of anything that prevents a view with subviews from being first responder.

commands/FBTextInputCommands.py
kastiglione: In chisel this idiom isn't common. I'd prefer we only guard if there's a reason for it, and even better if we could add a comment why that reason is.

commands/FBTextInputCommands.py
kastiglione: I see, thanks. How about `setinput`, `settext`? However, for existing accessibility commands, we use "a11y" naming. It's foreign, but we might want to be consistent with that? What do you think @gkassabli?

commands/FBTextInputCommands.py
tecknut: This command is intended to be used after selecting a text field/text view, making it first responder, then setting the text on it. 

With that in mind, if anything else is first responder at the time this command is run it will either do nothing, or set the text property on it. 

I took the happy path for this, but I could put checks in the `setTextInView` function which only set text on classes which respond to `setText:`.

commands/FBTextInputCommands.py
tecknut: I had started to write this in the accessibility command file. But I changed after thinking about the purpose of the command. It does use accessibility, but it's not the primary point of the commands.

commands/FBTextInputCommands.py
tecknut: Just looking back over the code, this branch of code is only executed when a view has no subviews.

commands/FBTextInputCommands.py
tecknut: I've been trying to replicate the error I had and haven't been able to. I'll push up changes with these checks taken out.

commands/FBTextInputCommands.py
kastiglione: > this branch of code is only executed when a view has no subviews.

Right, and I'm wondering about cases where some text input view has internal subviews. There's no guarantee that that a text input view will always have no subviews. I think the `isFirstResponder` check should go above the loop, and if it fails, then loop over subviews. This would make it breadth first, but I don't see a problem with that, do you?

tecknut: Nope, no issue. Pushed it up.

commands/FBTextInputCommands.py
tecknut: Pushed up `setinput` `settext`.

kastiglione:
I'm going to give this a day or two to see if @gkassabli has an feedback or thoughts about the accessibility part.

gkassabli:
Looks good to me

kastiglione:
Thanks!

tecknut:
Thanks guys !

PR #140, opened 2016-02-20
Added `pa11yi` to print accessibility identifiers of all views

bartoszj:
Hi,

While implementing ui tests we can use accessibility labels or accessibility identifiers. Currently `chisel` supports only commands for accessibility labels. I've added new command `pa11yi` to print accessibility identifiers of all views in the hierarchy.


kastiglione:
Thanks!

PR #141, opened 2016-02-26
Rename e{objc,swift} to {objc,swift}e

kastiglione:
In #121, two new chisel commands were added `eobjc` and `eswift`, for explicitly evaluating with a specific language. However this impacts the sanctity of the `e` command in lldb :stuck_out_tongue_winking_eye:. Sure I can from now on use `ex` where I previously used `e`, but we could also change the convention of how these commands are named.


kastiglione:
Thoughts @ultramiraculous @tomaskraina?

ultramiraculous:
Yeah...I had a coworker grumble about this the other day. My complaint here would be that I've personally gotten used to `es`/`eo`. 

Nobody wants to double down on this and just define `e` as a separate command...right? 😜

kastiglione:
> Nobody wants to double down on this and just define e as a separate command...right? 😜

hmm, I guess that's possible too…

PR #142, opened 2016-03-01
Use Carthage for building

kastiglione:
This change allows for building fat frameworks of Chisel using Carthage. 

1. Adds an empty `Cartfile`
2. Makes 'Chisel' Xcode scheme shared

Eventually the plan is for Chisel to make use of external libraries, so the `Cartfile` won't always be empty.

To build a fat framework, I ran the command:

```sh
carthage build --no-skip-current
```

To load the resulting framework at runtime, run this command in lldb:

```lldb
ex void* $chisel = (void*)dlopen("/path/to/Chisel.framework/Chisel", 1)
```

where `1` is the value of `RTLD_LAZY`. See `man dlopen`.

Once loaded, Chisel lldb commands can call into the Chisel.framework to do work that is best suited as "in-process". Other work that requires the lldb API can be performed by the python files.

### DIFF
--- a/Chisel.xcodeproj/xcshareddata/xcschemes/Chisel.xcscheme
+++ b/Chisel.xcodeproj/xcshareddata/xcschemes/Chisel.xcscheme
@@ -1,0 +1,99 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Scheme
+   LastUpgradeVersion = "0720"
+   version = "1.3">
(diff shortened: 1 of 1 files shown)


kastiglione:
cc @Gricha

kastiglione:
I'll update this description later to include steps for adding an external library.

Gricha:
:shipit:

PR #143, opened 2016-03-01
Introduce duplication function for Objective-C objects

Gricha:
This commit introduces very basic duplicate detection for Objective-C objects.


Chisel/CHLDuplicateDetection.m
```diff
@@ -0,0 +1,47 @@
+//
+//  CHLDuplicateDetection.m
+//  Chisel
+//
+//  Created by Greg Pstrucha on 3/1/16.
+//  Copyright © 2016 Facebook. All rights reserved.
+//
+
+#if __has_feature(objc_arc)
+#error CHLDuplicateDetection.m expects ARC to be disabled.
```
kastiglione: This file doesn't need ARC disabled. And actually, it needs ARC enabled because there are no `-release` calls, so the objects allocated in here are leaking.

Chisel/CHLDuplicateDetection.h
```diff
@@ -0,0 +1,18 @@
+//
+//  CHLDuplicateDetection.h
+//  Chisel
+//
+//  Copyright © 2016 Facebook. All rights reserved.
+//
+
+#ifndef CHLDuplicateDetection_h
+#define CHLDuplicateDetection_h
```
kastiglione: Will this file always be `#import`ed? If so, then these aren't necessary.

Chisel/CHLDuplicateDetection.h
```diff
@@ -0,0 +1,18 @@
...
+
+#import <Foundation/Foundation.h>
+
```
kastiglione: Wanna add `NS_ASSUME_NONNULL_BEGIN`?

Chisel/CHLDuplicateDetection.h
```diff
@@ -0,0 +1,18 @@
...
+/**
+ * Finds duplicates of objects in memory using `equalFunction` to compare.
+ */
+NSArray *CHLFindDuplicates(Class aCls, BOOL (^equalFunction)(id left, id right));
```
kastiglione: I dislike myself for asking, but can we either do `aClass` or `cls`?

Chisel/CHLDuplicateDetection.m
```diff
@@ -0,0 +1,47 @@
...
+#import "CHLObjectAllocations.h"
+
+#import <objc/runtime.h>
+
+NSArray *CHLFindDuplicates(Class aCls, BOOL (^equalFunction)(id left, id right)) {
+  NSMutableArray *objects = [NSMutableArray new];
+  
+  CHLEnumerateObjectsWithBlock(^(id object) {
+    if (object_getClass(object) == aCls)
+    [objects addObject:object];
```
kastiglione: Indentation is off. Can you use `{` and `}` around the `if`'s body?

ChiselTests/CHLDuplicateDetectionTests.m
```diff
@@ -0,0 +1,68 @@
...
+@end
+
+@interface CHLDuplicateDetectionTests : XCTestCase
+@end
+
+@implementation CHLDuplicateDetectionTests
+
+- (void)testDuplicateDetectionCanFindTwoCustomObjects
+{
+  @autoreleasepool {
```
kastiglione: Is an autorelease pool required?

Chisel/CHLDuplicateDetection.m
Gricha: Ah you are right, in the previous version I used MRR intentionally but end-solution definitely does not need it. I'll clean it up

Chisel/CHLDuplicateDetection.m
Gricha: Ha, good catch

ChiselTests/CHLDuplicateDetectionTests.m
Gricha: Oh I should clean it up, in the ideal world I thought it might be useful but also - the idea changed, not to mention these objects are retain cycles, so they get no releases :P

Chisel/CHLDuplicateDetection.m
Gricha: Ok I tested again, and I think it should be MRR after all. Reason for that is that object iterator will return anything that is an object, but cannot be retained. Example:

```objc
*** -[NSTaggedPointerStringCStringContainer retain] called, not supposed to happen
```

Chisel/CHLDuplicateDetection.m
```diff
@@ -0,0 +1,49 @@
...
+
+#import "CHLObjectAllocations.h"
+
+#import <objc/runtime.h>
+
+NSArray *CHLFindDuplicates(Class cls, BOOL (^equalFunction)(id left, id right)) {
+  NSMutableArray *objects = [NSMutableArray new];
+  
+  CHLEnumerateObjectsWithBlock(^(id object) {
+    if (object_getClass(object) == cls) {
```
kastiglione: I wonder if we should use `[object isKindOfClass:cls]` here? It will be pretty common to look for duplicates of `NSString`, `NSArray`, `NSDictionary`, etc, but those are all implemented as class clusters and as such are essentially abstract classes and no instances will be found. These collection objects will actually be internal subclasses, and I we definitely don't want to force callers to refer to internal subclasses for many reasons.

On the flip side, there might be cases where the caller wants to find duplicates based on exact class matching, not subclasses. I suspect this will be the less common case, so I would be fine with not supporting this case right now, and dealing with it later.

Gricha: I'm fine with both. I think using `isKindOfClass` for now might actually be better, since we might want to look at subclasses right away. Let me change that.

Chisel/CHLObjectAllocations.m
```diff
@@ -0,0 +1,49 @@
...
+    
+    if (![classes containsObject:class_]) {
+      return;
+    }
+    
+    if (range.size < class_getInstanceSize(class_)) {
+      return;
+    }
+    
+    if (range.size > class_getInstanceSize(class_) && object_getIndexedIvars(object) == NULL) {
```
kastiglione: This might not work, it will have to be tested or double checked. My question is around the nature of the `range` values returned by the malloc system. Does `range.size` indicate the class instance size, or does it indicate the full size of the block of memory that malloc returned.

Chisel/CHLObjectAllocations.m
```diff
@@ -0,0 +1,49 @@
...
+      return;
+    }
+    
+    block(object);
```
kastiglione: Another check to perform before calling the block, is to check if the object is being deallocated, presumably by calling `-allowsWeakReference`.

Chisel/CHLObjectAllocations.m
```diff
@@ -0,0 +1,49 @@
...
+  }
+  free(classList);
+  
+  CHLEnumerateAllocationsWithBlock(^(vm_range_t range) {
+    if (range.size < sizeof(struct objc_object)) {
+      return;
+    }
+    
+    id object = (id)range.address;
+    Class class_ = object_getClass(object);
```
kastiglione: I wonder what all possible outcomes are when `object` is not actually an objc object. At this point, all that is known is that the memory is large enough, but a lot of allocations will pass that test. Will `object_getClass` return `nil` if it's given something that's not an objc object, or could it assert and crash? If the latter, this code will need it's own `getClass` implementation.

PR #150, opened 2016-04-19
Add present and dismiss commands for UIViewController

alanzeino:
How I tested:

```
(lldb) expr UIViewController *$avc = (UIViewController *)[[UIViewController alloc] init]
(lldb) dismiss $avc
  File "/Users/alan/repositories/github/chisel/fblldb.py", line 79, in runCommand
    command.run(args, options)
  File "/Users/alan/repositories/github/chisel/commands/FBDisplayCommands.py", line 235, in run
    viewControllerHelpers.dismissViewController(args[0])
  File "/Users/alan/repositories/github/chisel/fblldbviewcontrollerhelpers.py", line 37, in dismissViewController
    raise Exception('Argument must be presented')
Exception: Argument must be presented
(lldb) present $avc
(lldb) c
Process 8420 resuming
(lldb) present $avc
  File "/Users/alan/repositories/github/chisel/fblldb.py", line 79, in runCommand
    command.run(args, options)
  File "/Users/alan/repositories/github/chisel/commands/FBDisplayCommands.py", line 221, in run
    viewControllerHelpers.presentViewController(args[0])
  File "/Users/alan/repositories/github/chisel/fblldbviewcontrollerhelpers.py", line 24, in presentViewController
    raise Exception('Argument is already presented')
Exception: Argument is already presented
(lldb) dismiss $avc
(lldb) c
```

### DIFF
--- a/fblldbviewcontrollerhelpers.py
+++ b/fblldbviewcontrollerhelpers.py
@@ -12,6 +12,32 @@ import lldb
 import fblldbbase as fb
 import fblldbobjcruntimehelpers as runtimeHelpers
 
+def presentViewController(viewController):
+  vc = '(%s)' % (viewController)
+  
+  if fb.evaluateBooleanExpression('%s != nil && ((BOOL)[(id)%s isKindOfClass:(Class)[UIViewController class]])' % (vc, vc)):
+    notPresented = fb.evaluateBooleanExpression('[%s presentingViewController] == nil' % vc)
+    
+    if notPresented:
+      lldb.debugger.HandleCommand('expr (void)[[[[UIApplication sharedApplication] keyWindow] rootViewController] presentViewController:%s animated:YES completion:nil]' % vc)
+    else:
+      raise Exception('Argument is already presented')
+  else:
+    raise Exception('Argument must be a UIViewController')
+
+def dismissViewController(viewController):
+  vc = '(%s)' % (viewController)
+  
+  if fb.evaluateBooleanExpression('%s != nil && ((BOOL)[(id)%s isKindOfClass:(Class)[UIViewController class]])' % (vc, vc)):
+    isPresented = fb.evaluateBooleanExpression('[%s presentingViewController] != nil' % vc)
+    
+    if isPresented:
+      lldb.debugger.HandleCommand('expr (void)[(UIViewController *)%s dismissViewControllerAnimated:YES completion:nil]' % vc)
+    else:
+      raise Exception('Argument must be presented')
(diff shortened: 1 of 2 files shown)


fblldbviewcontrollerhelpers.py
```diff
@@ -12,6 +12,32 @@
 import fblldbbase as fb
 import fblldbobjcruntimehelpers as runtimeHelpers
 
+def presentViewController(viewController):
+  vc = '(%s)' % (viewController)
+  
+  if fb.evaluateBooleanExpression('%s != nil && ((id)[(id)%s isKindOfClass:(Class)[UIViewController class]])' % (vc, vc)):
```
kastiglione: The `(id)` cast after the `&&` should be `(BOOL)`.

fblldbviewcontrollerhelpers.py
```diff
@@ -12,6 +12,32 @@
...
+      lldb.debugger.HandleCommand('expr (void)[[[[UIApplication sharedApplication] keyWindow] rootViewController] presentViewController:%s animated:YES completion:nil]' % vc)
+    else:
+      raise Exception('Argument is already presented')
+  else:
+    raise Exception('Argument must be a UIViewController')
+
+def dismissViewController(viewController):
+  vc = '(%s)' % (viewController)
+  
+  if fb.evaluateBooleanExpression('%s != nil && ((id)[(id)%s isKindOfClass:(Class)[UIViewController class]])' % (vc, vc)):
```
kastiglione: The `(id)` cast after the `&&` should be `(BOOL)`.

alanzeino:
Updated!

kastiglione:
🚢

kastiglione:
Thanks Xzibit

joshavant:
Any thoughts on whether it's presently possible to push a view controller from a specific `UINavigationController` instance vs presenting it via `UIViewController.present()`?

alanzeino:
As in, to push it onto a specific UINavigationController?

It's doable, but the question is do we want a command that takes two arguments because I can't see it working without something like this:

```
push $nc $vc
pop $vc
```

(where $nc is the navigation controller and $vc is the view controller)

`pop` can probably omit the $nc I guess.

joshavant:
Yes.

The use case here is that I'm trying to present a specific View Controller for debugging, as close as possible to how it would organically be presented in the app.

If this is out-of-scope for the framework, perhaps the solution here is dropping into the Swift REPL to accomplish this.

PR #152, opened 2016-04-28
Add 'pbundlepath' Print application's bundle directory path

Ashton-W:
Very similar to the existing `pdocspath`, this new command `pbundlepath` prints the path to the app bundle. Basically just `[[NSBundle mainBundle] bundlePath]` but wrapped in a nice chisel command.

### DIFF
--- a/commands/FBPrintCommands.py
+++ b/commands/FBPrintCommands.py
@@ -31,6 +31,7 @@ def lldbcommands():
     FBPrintInstanceVariable(),
     FBPrintKeyPath(),
     FBPrintApplicationDocumentsPath(),
+    FBPrintApplicationBundlePath(),
     FBPrintData(),
     FBPrintTargetActions(),
     FBPrintJSON(),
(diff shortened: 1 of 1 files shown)


kastiglione:
Thanks. I feel like it would be good to have just one command for "all" things path related. For example:

Print docs path:
`ppath -d`

Print bundle path:
`ppath -b`

or, something like that.

Although I said "all", for now we can handle just the two paths, and add any other paths down the road. By default, `ppath` could print both the doc and bundle path, but wouldn't put anything in the pasteboard I guess? People can easily create their own `pdocspath` or `pbundlepath` aliases in their `.lldbinit`.

Thoughts?

Ashton-W:
I originally wrote and shared this as a simple alias for use in an `.lldbinit` file and it was suggested to add t to chisel.

I can't think of any others paths that aren't closely relative to either of these two directories - so I don't see anyone having a good case for adding more ppath type commands.

What do you think?

Ashton-W:
I do like the idea of having options for these commands as shortcuts to particular directories of interest eg: Caches

Ashton-W:
Ha, in saying that I do see how two commands for the different sets of interesting app directories could be confusing - even if they are grouped by which path they are relative to. :)


(Sorry no edit on mobile github)

kastiglione:
it sounds like you're agreeing with the idea of one unified command, but I'm not 100% sure I'm interpreting your comments correctly…

Ashton-W:
I'll try creating a unified command

kastiglione:
Ideally we wouldn't introduce a command only to remove it. If the unified command isn't going to happen any time soon, then yep I'll merge this. If you're aiming to soon work on the unified command, then I would hold off on this pull request.

Ashton-W:
The unified command is a bit more complicated, and I think it should be opened up to other contributors anyway.

kastiglione:
> opened up to other contributors

Chisel contributions of awesome commands, but not many contributors do refactoring.

If these commands get merged, I'll @mention you when to give you the heads up 😉

Thanks!

PR #154, opened 2016-05-06
Provide a help string when registering commands

kastiglione:
I don't know when it was first added, but `command script add` has a `--help` flag, which defines the help text to show in the output of the `help` command.

This change calls the description method on each command, extracts the first line, and then passes that as the help string when registering the command.

Example output:

```
pa11y        -- Print accessibility labels of all views in hierarchy of <aView>
pa11yi       -- Print accessibility identifiers of all views in hierarchy of <aView>
pactions     -- Print the actions and targets of a control.
paltrace     -- Print the Auto Layout trace for the given view. Defaults to the key window.
panim        -- Prints if the code is currently execution with a UIView animation block.
pbcopy       -- Print object and copy output to clipboard
```

### DIFF
--- a/fblldb.py
+++ b/fblldb.py
@@ -41,7 +41,7 @@ def loadCommandsInDirectory(commandsDirectory):
 def loadCommand(module, command, directory, filename, extension):
   func = makeRunCommand(command, os.path.join(directory, filename + extension))
   name = command.name()
-  helpText = command.description().split('\n', 1)[0] # first line of description
+  helpText = command.description().splitlines()[0] # first line of description
 
   key = filename + '_' + name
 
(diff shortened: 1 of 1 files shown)


fblldb.py
```diff
@@ -41,6 +41,7 @@ def loadCommandsInDirectory(commandsDirectory):
 def loadCommand(module, command, directory, filename, extension):
   func = makeRunCommand(command, os.path.join(directory, filename + extension))
   name = command.name()
+  helpText = command.description().split('\n', 1)[0] # first line of description
```
modocache: You could also use `str.splitlines()` here instead, unless the `maxsplit=1` here is important to you.

kastiglione: That's definitely more readable. The use of `maxsplit` was to avoid doing more work than necessary.

modocache:
LGTM!

kastiglione:
✨

Issue #15, opened 2014-03-07
Create test suite


kastiglione:
Let's build a test suite for Chisel. I'm thinking integration tests.
- [ ] Build an app for the test harness.
- [ ] Build test runner
- [ ] ~~Profit~~ Write tests

Thoughts?

facebook-github-bot-3:
Thank you for reporting this issue and appreciate your patience. We've  notified the core team for an update on this issue. We're looking for a response within the next 30 days or the issue may be closed.

PR #155, opened 2016-05-07
Expression refactor

kastiglione:
This pull request refactors how chisel handles evaluation. There are two improvements:

1. Evaluate user input in the currently active language, presumably either swift or objc
2. Perform evaluation of generated code using lldb APIs, instead of workarounds like `eobjc`

Currently, the first issue is an existing bug in chisel. When stopped at some swift code, the user will rightly expect to call chisel commands with code expressions written in swift, not objc.

The second issue is more of preference. I would like to remove the `eobjc`, `poobjc`, `eswift`, and `poswift` commands.

Todo:

- [x] Refactor `FBDisplayCommands.py`
- [x] Refactor `FBAutoLayoutCommands.py`
- [x] Refactor `FBAccessibilityCommands.py`
- [x] Refactor `FBComponentCommands.py`
- [x] Refactor `FBInvocationCommands.py`
- [x] Refactor `FBVisualizationCommands.py`
- [x] Refactor `fblldbviewhelpers.py`
- [x] Remove use of `poobjc`
- [x] Remove ~~`poobjc`~~, `eobjc`, `eswift`, and `poswift`

Extras: I ended up doing some more refactoring, particularly by adding `describeObject`, which gives the output that `po` would print, and `evaluateEffect`, which is a special case method for calling code that has  a `void` result type.


kastiglione:
@ultramiraculous Do you have any thoughts on the removal of `poobjc`, `poswift`, etc? There's a mix of motivations in this pull request, but I'm curious about your thoughts from the user's perspective. Internally, the uses of those commands has been replaced with calls into the lldb API. This should be invisible to the user. However, I'm wondering how many users are heavily using these commands directly. I've got to go for now, but I can elaborate in more detail this weekend.

ultramiraculous:
The rest of this sounds great, but I do actually personally use `poobjc` constantly day to day. It's pretty useful to be able to just work with ObjC objects in an ObjC context when the Swift parts of lldb aren't feeling helpful/are flat out broken. It's decent escape hatch when you've got the memory address of an object and don't want to bother finding a stack frame that evaluates expressions as ObjC++.

ultramiraculous:
That said, I'm 100% willing to believe the list of heavy users might be me and a coworker. If you have a good argument I'm not looking to die on this hill/be "that guy" over my pet use case 😛

kastiglione:
That sounds like a pretty convincing reason. I'd be ok to keep `poobjc` for that reason.

One of the things I've yet to find a balance for is that some chisel commands are simple enough that they could be defined as a `command alias` in `.lldbinit`, and I mention this because  `poobjc` is one of those. If something can be a `command alias`, maybe it should be? This is just me thinking aloud.

I'll add `poobjc` back in. Thanks for the feedback. The escape hatch makes a lot of sense, and now I'm wondering whether all commands need to be able to handle objc input.

kastiglione:
I've tested a bunch of the affected commands manually, and they work for me. One day we'll have #15 :trollface:

kastiglione:
I figure I'll merge this if there are no reviewers. cc @facebook/team-chisel-restored

PR #167, opened 2016-08-15
Add class to recognize Objective-C objects in memory

kastiglione:
This is a first pass at recognizing memory allocations that are Objective-C objects. There are probably better ways to do this, suggestions and pointers welcome!

### DIFF
--- a/Chisel/CHLObjCObjectRecognizer.mm
+++ b/Chisel/CHLObjCObjectRecognizer.mm
@@ -5,6 +5,10 @@
 //  Copyright © 2016 Facebook. All rights reserved.
 //
 
+#if __has_feature(objc_arc)
+#error CHLObjCObjectRecognizer.m expects ARC to be disabled.
+#endif
+
 #import "CHLObjCObjectRecognizer.h"
 
 #include <unordered_map>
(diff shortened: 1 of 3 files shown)


Chisel/CHLObjCObjectRecognizer.mm
```diff
@@ -0,0 +1,109 @@
...
+
+- (BOOL)isStandardAllocUsedByClass:(Class)cls
+{
+  const auto match = _cache.find(cls);
+  if (match != _cache.cend()) {
+    return match->second;
+  }
+
+  const Class metaClass = object_getClass(cls);
+  const bool usesStandardAlloc =
```
zats: What about `NSProxy` subclasses?

kastiglione: 😛

This pull request is old, I'll close it. It is a good question though. My first thought is that it's ok to ignore `NSProxy`, because proxies are semantically not the type of class that would allocate much, especially not variable length objects like arrays and other collections.

zats: sorry, I was wondering around the repo 🙂 
you are right, I saw proxies mostly being used for mocking in testing frameworks, not sure if it's a production use-case

PR #168, opened 2016-08-26
Evaluate with SetTrapExceptions set to False

kastiglione:
With this change, code that happens to throw and catch an internal exception will now successfully evaluate to completion. Without this, no result would be returned, and instead this error would be printed to the console:

    error: Execution was interrupted, reason: internal ObjC exception breakpoint(-3)..


kastiglione:
cc @Gricha

kastiglione:
#yolo

PR #173, opened 2016-11-08
Add pivars command to FBClassDump.py

longv2go:
`pivars` command call the `class_copyIvarList` to get all ivars of an Class and print them out. It would show the offset, type, name and value if need.

```
(lldb) pivars -r -n child
NSObject:
    [0x0] Class isa

father:
    [0x4] NSString * father_var
    [0x8] NSString * faterh_var2

child:
    [0xc] {teststr="a"i"b"l} straaa
    [0x14] NSString * c_var1
    [0x18] NSString * _c_var2

```

### DIFF
--- a/commands/FBClassDump.py
+++ b/commands/FBClassDump.py
@@ -8,7 +8,8 @@ def lldbcommands():
   return [
     FBPrintMethods(),
     FBPrintProperties(),
-    FBPrintBlock()
+    FBPrintBlock(),
+    FBPrintIvars()  
   ]
 
 class FBPrintMethods(fb.FBCommand):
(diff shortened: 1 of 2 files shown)


kastiglione:
Thanks, and sorry for the late reply. This looks good, but I'm also wondering if we should consolidate `pinternals` and `pivar`? Should we discuss that now, or leave that for another time.

longv2go:
@kastiglione I notice that too. `pinternals`  can not show the ivar\`s offset, so I wrote this command. Sometimes I need the offset to inspect the object\`s memory layout,  and ```pivar``` need the ivar`s name, no name no working.

kastiglione:
Related, a command I came across last week is:

```
language objc class-table dump -v <regular-expression>
```

This prints ivars and their offsets, but it also prints methods too, which adds noise if you just want structure information.

longv2go:
@kastiglione I think we should to merge the pivars, pmethods, pproperties to one command. 
```
class-dump <options> <arg>
```
if the arg is an instance of some class, it can show the ivar`s value of that instance. But there is more work to decode the struct infomation.

kastiglione:
@longv2go I'm following up on outstanding pull requests. Sorry for the long silence. Note that since this time, a `-a` flag was added to `pinternals` which calls the debugging helper `_ivarDescription`.

PR #175, opened 2016-11-21
Remove outdated CKComponentAction type summary

kastiglione:
`CKComponentAction` is no longer a `typedef` for `SEL`. See https://github.com/facebook/componentkit/pull/676

cc @ocrickard @eczarny


ocrickard:
Ahh, nice. I need to update our docs as well. I've been busy with other things. Will do this week though.

PR #179, opened 2016-12-11
Add delay command

zats:
This adds a delayed execution of commands for lldb

## Reasoning

Often when debugging involved user interaction or interactive transitions in simulator, it takes some time to setup the interaction I need. Once setup, it is not unheard of that I can't switch away from simulator: if gesture-driven animation is in flight, switching away from simulator might interrupt it.

## Usage

This is why I found it useful to do following:

`$ zzz 3 "process interrupt"` (== `$ zzz 3`)
`$ zzz 3 pcomponents` 
etc… 

This restarts execution, gives me 3 seconds to setup my interaction, and executes specified command afterwards.

My python is quite rusty, any feedback is highly appreciated!

### DIFF
--- a/commands/FBDelay.py
+++ b/commands/FBDelay.py
@@ -1,0 +1,37 @@
+#!/usr/bin/python
+from threading import Timer
+import fblldbbase as fb
+import fblldbobjcruntimehelpers as runtimeHelpers
+import lldb
+import string
+
+
+def lldbcommands():
+  return [
+    FBDelay()
+  ]
+
+class FBDelay(fb.FBCommand):
+  def name(self):
+    return 'zzz'
+
+  def description(self):
+    return 'Executes specified lldb command after delay.'
+
+  def args(self):
+    return [
+      fb.FBCommandArgument(arg='delay in seconds', type='float', help='time to wait before executing specified command'),
+      fb.FBCommandArgument(arg='lldb command', type='string', help='another lldb command to execute after specified delay', default='process interrupt')
+    ]
+
+  def run(self, arguments, options):
+    lldb.debugger.SetAsync(True)
+    lldb.debugger.HandleCommand('process continue')
+    delay = float(arguments[0])
+    command = str(arguments[1])
+    t = Timer(delay, lambda: self.runDelayed(command))
+    t.start()
+
+  def runDelayed(self, command):
+    lldb.debugger.HandleCommand('process interrupt')
+    lldb.debugger.HandleCommand(command)


zats:
Oh and if there is a built-in way of doing this, let me know 😆

commands/FBDelay.py
```diff
@@ -0,0 +1,40 @@
...
+
+class FBDelay(fb.FBCommand):
+  def name(self):
+    return 'zzz'
+
+  def description(self):
+    return 'Executes specified lldb command after delay.'
+
+  def options(self):
+    return []
```
kastiglione: I think this can be left out.

commands/FBDelay.py
```diff
@@ -0,0 +1,40 @@
...
+
+  def args(self):
+    return [
+      fb.FBCommandArgument(arg='delay in seconds', type='float', help='time to wait before executing specified command'),
+      fb.FBCommandArgument(arg='lldb command', type='string', help='another lldb command to execute after specified delay', default='process interrupt')
+    ]
+
+  def run(self, arguments, options):
+    lldb.debugger.SetAsync(True)
+    lldb.debugger.HandleCommand('continue')
```
kastiglione: There is a `Continue()` function that can be called directly, but given the way Chisel is currently architected, doing it this way is cleaner. Wanna change this to `'process continue'` to match the `'process interrupt'` used below?

commands/FBDelay.py
```diff
@@ -0,0 +1,40 @@
...
+    return []
+
+  def args(self):
+    return [
+      fb.FBCommandArgument(arg='delay in seconds', type='float', help='time to wait before executing specified command'),
+      fb.FBCommandArgument(arg='lldb command', type='string', help='another lldb command to execute after specified delay', default='process interrupt')
+    ]
+
+  def run(self, arguments, options):
+    lldb.debugger.SetAsync(True)
```
kastiglione: Should this code save the current value of async and set it back after?

kastiglione:
Looks legit, amazing name too. I want to give Chisel some aliasing features, in which case this could be aliased to `delay` as well. I have just a couple comments.

kastiglione:
Ideally it wouldn't be necessary to put the command in quotes, but that can be a fix for another day.

commands/FBDelay.py
```diff
@@ -0,0 +1,40 @@
...
+    lldb.debugger.HandleCommand('continue')
+    delay = float(arguments[0])
+    command = str(arguments[1])
+    t = Timer(delay, lambda: self.runDelayed(command))
+    t.start()
+
+  def runDelayed(self, command):
+    lldb.debugger.HandleCommand('process interrupt')
+    lldb.debugger.HandleCommand(command)
```
kastiglione: Does lldb silently accept an empty command? I ask because the description says `zzz 3 "process interrupt"` is equivalent to just `zzz 3`, so I'm presuming this means lldb doesn't error on empty commands.

kastiglione:
An interesting addition to this command some day is give it a "breakpoint equivalent" (function, line number, etc) and then the given command is run _either_ when the breakpoint is hit, or when the delay/timeout happens.

commands/FBDelay.py
zats: I actually set a default value on the command option, it probably got lost because I didn't break it at certain line length (is there a style guide you'd like to follow here?)
The default parameter is `, default='process interrupt'`
Should I add a check for an empty command and replace it with default value?

commands/FBDelay.py
zats: seems like I didn't understand the meaning of `SetAsync()`. 
From [documentation](http://lldb.llvm.org/python_reference/lldb.SBDebugger-class.html) I thought that it detaches execution of the following command (`continue` in this case) allowing function to be evaluated even after we continue the execution.
Did I misunderstand it? Or maybe I'm just missing the point of your comment

kastiglione: I think that's right, but I didn't think its affect was for only one command, I'm thinking it stays in async mode until something calls `SetAsync(False)`. If that is the case this command should change it back to the original value after it's done.

commands/FBDelay.py
kastiglione: 😎 I did overlook the default.

Given the declared default, don't feel strongly either way about handling an empty command.

kastiglione:
Re the breakpoint idea, let me illustrate

```
delay "-[Foo didDoThing]" 3 pcomponents
```

Basically it's like a expiring breakpoint. Within the next three seconds, if the method is called, run pcomponents, otherwise call pcomponents when time runs out.

I guess the above is just three commands:

```
tbreak -[Foo doThatThing]
zzz 3 "break dis <bp>"
zzz 3 pcomponents
```

Which raises a question, can two or more `zzz` commands be run like that?

zats:
Hmm that's an interesting idea, but I'd probably extract it into a separate command or as an option to the existing one. Regardless seems like a material for another pull request. I'll make sure to fix other comments in the meantime

commands/FBDelay.py
zats: I actually don't see it being used like that in [other cases](https://github.com/facebook/chisel/search?utf8=✓&q=lldb.debugger.SetAsync&type=Code) we either enable reset async conditionally or don't do it at all in [FBFindCommands.py](https://github.com/facebook/chisel/blob/c3d7e95ebdbae03e246bcb56e19da77fddea5e16/commands/FBFindCommands.py#L119)

I don't mind resetting if you are certain we should do it. 
Sadly, no documentation, and I don't feel like lldb spelunking right now 😳

kastiglione: I spelunked last night, and the main thing is that the async setting decides whether `Resume()` or `ResumeSynchronous()` is called. I didn't write any code though which would really answer the question. I don't know if those other cases really confirm it either way.

Since it must work as is, we can leave this for another day 👍

kastiglione:
Looks ready, you don't have any changes planned right?

zats:
Oh sorry, lost track of this PR. I didn't plan anything unless I forgot to address any of your comments?

kastiglione:
Thanks!

PR #183, opened 2017-03-03
Add support for CKComponent passed into pcomponents

zats:
Sometime I pass a component reference into `pcomponents` by mistake. This should allow it not to explode

### DIFF
--- a/commands/FBComponentCommands.py
+++ b/commands/FBComponentCommands.py
@@ -4,6 +4,7 @@ import os
 
 import lldb
 import fblldbbase as fb
+import fblldbviewhelpers as viewHelpers
 
 def lldbcommands():
   return [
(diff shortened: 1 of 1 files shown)


commands/FBComponentCommands.py
```diff
@@ -50,13 +51,17 @@ def options(self):
...
+    return [ fb.FBCommandArgument(arg='aView', type='UIView* or CKComponent*', help='The view or component from which the search for components begins.', default='(id)[[UIApplication sharedApplication] keyWindow]') ]
 
   def run(self, arguments, options):
     upwards = 'YES' if options.upwards else 'NO'
     showViews = 'YES' if options.showViews == 'YES' else 'NO'
 
     view = fb.evaluateInputExpression(arguments[0])
+    if not viewHelpers.isView(view):
+        # assume it's a CKComponent
+        view = fb.evaluateExpression('((CKComponent *)%s).viewContext.view' % view)
```
kastiglione: If you do more changes to `pcomponents` in the future, it'd be nice to check that it's a `CKComponent`.

zats: thanks for taking a look and agreed! 
we need to move to that approach when we can write plugins in something more "objective" and something more "C" to simplify dealing with plugins

PR #187, opened 2017-03-18
Add open sandbox command

jianzhi2010:
Adding 1 command : osand
This command open the Simulator sandbox folder of the app in Finder.

### DIFF
--- a/commands/FBOpenCommands.py
+++ b/commands/FBOpenCommands.py
@@ -1,0 +1,23 @@
+#!/usr/bin/python
+
+import os
+
+import lldb
+import fblldbbase as fb
+
+def lldbcommands():
+  return [ PrintKeyWindowLevel() ]
+
+class PrintKeyWindowLevel(fb.FBCommand):
+  def name(self):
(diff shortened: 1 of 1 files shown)


commands/FBOpenCommands.py
```diff
@@ -0,0 +1,23 @@
...
+
+import os
+
+import lldb
+import fblldbbase as fb
+
+def lldbcommands():
+  return [ PrintKeyWindowLevel() ]
+
+class PrintKeyWindowLevel(fb.FBCommand):
```
kastiglione: `class OpenSandboxBox`?

kastiglione:
I missed this pull request when it came in.

I just compared this command to `pdocspath`, and the results are the same.

`pdocspath` calls `NSSearchPathForDirectoriesInDomains(NSDocumentDirectory, …)`, and at least for Simulator apps this is the same as calling `NSHomeDirectory()`.

Thanks for the pull request, but since this is already provided by Chisel, I will close it.

PR #189, opened 2017-04-13
Force start accessibility for `pa11yi` command

bartoszj:
On the `pa11yi` command an error is returned:

```
UIWindow (id)[[UIApplication sharedApplication] keyWindow]
error: error: Execution was interrupted, reason: Attempted to dereference an invalid ObjC Object or send it an unrecognized selector.
error: error: use of undeclared identifier 'None'
```

The `pa11y` commands works without any problem. The reason of this error is not started Accessibility Server.


kastiglione:
thank you!

PR #192, opened 2017-05-18
Added `xdebug`, `xtree`, `xobject` and `xnoid` commands

bartoszj:
This PR adds four new commands to help with Xcode UI tests. All commands are working with Objective-C and Swift code.

#### `xdebug`

This command is a shortcut for `XCUIApplication().debugDescription` (or `[[[XCUIApplication alloc] init] debugDescription]` for Objective-C). It accepts an optional argument which an instance of the `XCUIElement` type which will be used instead of `XCUIApplication`. In addition, for the Swift variant, all `\n` characters are normalized to new line characters.

#### `xtree`

Prints tree of `XCUIElement` objects of the `XCUIApplication` object (it can be configured by an optional argument like in the `xdebug` command). Output is similar to `XCUIApplication().debugDescription` but it doesn't contain by default "unnecessary" information like frame, traits or pointer value. That additional information can be turned on by using options `--frame`, `--trait` and `--pointer`.

Example output:

```
Application: label: 'test'
 | Window: MainWindow
 |  | Other: identifier: 'mainView'
 |  |  | Button: identifier: 'buttonWithId', label: 'Button with identifier'
 |  |  | Button: label: 'Button without identifier'
 |  |  | Button: identifier: 'goButton', label: 'GO!'
 | Window: 
 |  | StatusBar: 
 |  |  | Other: 
 |  |  | Other: 
 |  |  |  | Other: 
 |  |  |  | Other: label: '3 of 3 Wi-Fi bars', value: 'SSID'
 |  |  |  | Other: label: '1:51 PM'
 |  |  |  | Other: label: '-100% battery power'
```

#### `xnoid`

This command can be used to find potential views with missing identifiers. It prints all objects which have set label but doesn't have the identifier. Output is similar to `xtree` command but in addition, the `StatusBar` is not presented (which can be changed with `--status-bar` option).

Example output:

```
Application: label: 'test'
 | Window: MainWindow
 |  | Other: identifier: 'mainView'
 |  |  | Button: label: 'Button without identifier'
```

#### `xobject`

Prints detailed information about the `XCUIElement` object.

Example output:

```
(lldb) xobject button
    t =   103.10s     Snapshot accessibility hierarchy for com.pgs-soft.test
    t =   103.46s     Find: Descendants matching type Button
    t =   103.46s     Find: Elements matching predicate '"goButton" IN identifiers'
Pointer: 0x608000365880
Type: Button
Depth: 3
Traits: Button (0x200000001)
Frame: {{145.0, 269.0}, {30.0, 30.0}}
Visible frame: {{145.0, 269.0}, {30.0, 30.0}}
Identifier: 'goButton'
Label: 'GO!'
Title: ''
Value: ''
Placeholder: ''
Enabled: True
Selected: False
Main Window: False
Keyboard focus: False
Focus: False
Generation: 3
Horizontal size class: Compact
Vertical size class: Regular
TouchBar element: False
TouchBar top level element: False
Unique Objective-C: (
[[XCUIApplication alloc] init].buttons[@"goButton"],
[[XCUIApplication alloc] init].buttons[@"GO!"],
[[XCUIApplication alloc] init].otherElements[@"mainView"].buttons[@"goButton"],
[[XCUIApplication alloc] init].otherElements[@"mainView"].buttons[@"GO!"]
)
Unique Swift: (
XCUIApplication().buttons["goButton"],
XCUIApplication().buttons["GO!"],
XCUIApplication().otherElements["mainView"].buttons["goButton"],
XCUIApplication().otherElements["mainView"].buttons["GO!"]
)
Suggested hit points: (
NSPoint: {160, 284}
)
```

### DIFF
--- a/commands/FBXCTestCommands.py
+++ b/commands/FBXCTestCommands.py
@@ -1,0 +1,1390 @@
+#!/usr/bin/python
+
+# Copyright (c) 2017, Facebook, Inc.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree. An additional grant
+# of patent rights can be found in the PATENTS file in the same directory.
+
+import lldb
+import fblldbbase as fb
+import re
+
+NOT_FOUND = 0xffffffff  # UINT32_MAX
+
+
+def lldbcommands():
+  return [
+    FBXCPrintDebugDescription(),
+    FBXCPrintTree(),
+    FBXCPrintObject(),
+    FBXCNoId(),
+  ]
+
+
+class FBXCPrintDebugDescription(fb.FBCommand):
+  def name(self):
+    return 'xdebug'
+
+  def description(self):
+    return 'Print debug description the XCUIElement in human readable format.'
+
+  def args(self):
+    return [fb.FBCommandArgument(arg='element', type='XCUIElement*', help='The element to print debug description.', default='__default__')]
+
+  def run(self, arguments, options):
+    element = arguments[0]
+    language = fb.currentLanguage()
+
+    if element == '__default__':
+      element = 'XCUIApplication()' if language == lldb.eLanguageTypeSwift else '(XCUIApplication *)[[XCUIApplication alloc] init]'
+
+    if language == lldb.eLanguageTypeSwift:
+      print fb.evaluateExpressionValue("{}.debugDescription".format(element), language=language) \
+        .GetObjectDescription() \
+        .replace("\\n", "\n") \
+        .replace("\\'", "'") \
+        .strip(' "\n\t')
+    else:
+      print fb.evaluateExpressionValue("[{} debugDescription]".format(element)).GetObjectDescription()
+
+
+class FBXCPrintTree(fb.FBCommand):
+  def name(self):
+    return "xtree"
+
+  def description(self):
+    return "Print XCUIElement subtree."
+
+  def args(self):
+    return [fb.FBCommandArgument(arg="element", type="XCUIElement*", help="The element to print tree.", default="__default__")]
+
+  def options(self):
+    return [
+      fb.FBCommandArgument(arg='pointer', short='-p', long='--pointer', type='BOOL', boolean=True, default=False, help='Print pointers'),
+      fb.FBCommandArgument(arg='trait', short='-t', long='--traits', type='BOOL', boolean=True, default=False, help='Print traits'),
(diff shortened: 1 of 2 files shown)


kastiglione:
Thanks for submitting this!

I've only taken a quick look, I'll review it soon.

I do have one initial question: Why have both swift and objc expressions, why not use objc everywhere? I suspect I know what the answer is, but maybe I don't.

bartoszj:
Hi,

All commands accept an expression as an argument. So, in a case of the `xdebug` user can provide Swift or Objective-C expression depending on the source code, for example:

```swift
xdebug userPage.nameLabel()
```
After evaluating the expression in the Swift or the Objective-C those commands are getting a pointer to the evaluated object and perform all tasks in Objective-C for simplicity. So both expressions are used only at the beginning to handle user provided data.

kastiglione:
Right. The technique (trick) I've used is to put the user input in backquotes. For example:

```
pviews self.view
```

doesn't work if called from a view controller written in swift, but this does:

```
pviews `self.view`
```

Many chisel commands are like this, they currently don't always work from swift. I was thinking that commands could add backticks around the user input to resolve this. I haven't done any work to update the existing chisel commands, but you might want to consider it for these new commands.

On the flip side, maybe there's some value evaluating swift expressions in swift, I haven't given it much thought.

bartoszj:
Hi,

I didn't know about `` ` ``, thanks for the hint. But unfortunately, it doesn't work from Chisel/Python.

When I execute command directly from LLDB like:

```
expression --language objc++ -O -- [(XCUIElement *)`XCUIApplication()` debugDescription]
```

it works. But when I execute exactly the same expression from Python:

```python
print fb.evaluateExpressionValue("[(XCUIElement *)`XCUIApplication()` debugDescription]").GetObjectDescription()
```
I'm getting an error:

```
error: error: expected expression
None
```

I couldn't find any option in the [`SBExpressionOptions `](https://lldb.llvm.org/cpp_reference/html/classlldb_1_1SBExpressionOptions.html) which would fix this issue and I've no idea how to get around this problem.

bartoszj:
Hi,

Are you content with my PR or should I change something?

kastiglione:
Sorry for the delay, thank you the pull request. I'll try to review today/tomorrow.

kastiglione:
Looks good to me!

kastiglione:
I should clarify, I went over it pretty lightly so I wouldn't have caught any bugs with my eye. Thanks again!

kastiglione:
Note to self, maybe these commands should be aded to the README.

PR #197, opened 2017-07-23
Add findinstances, and new support framework in Chisel.xcodeproj

kastiglione:
This adds a new command, `findinstances`, which is implemented mostly in native code and so comes with Chisel.xcodeproj which contains a new Chisel framework. Future Chisel commands can implement the heavy lifing this way too.

The `findinstances` command scans the heap using available iOS/macOS malloc API. For each allocation, a number of heuristics are peformed to identify likely Objective-C instances. The heuristics do not call methods on the objects, instead relying only on objc runtime functions to passively match the instance based on its class metadata. This avoids allocations and stateful side effects in the objc runtime.

After this first pass, the candidate objects go through a second pass that checks if they match against an optional `NSPredicate`. If there's no predicate, the object is printed with minimal information. If there is a predicate, and the object passes the predicate, then the object will be printed out with more detail, specifically the detail queried in the predicate.

Some examples of `findinstances`:

```
findinstances UIView
findinstances *UIView
findinstances UIScrollViewDelegate
findinstances UIView window == nil || hidden == true || alpha == 0 || layer.bounds.#size.width == 0 ||  layer.bounds.#size.height == 0 
findinstances UIView subviews.@count == 0
findinstances NSDictionary any @allKeys beginswith 'perf_'
findinstances NSArray @count > 100
```

### DIFF
--- a/Chisel/Chisel.xcodeproj/project.pbxproj
+++ b/Chisel/Chisel.xcodeproj/project.pbxproj
@@ -1,0 +1,439 @@
+// !$*UTF8*$!
+{
+	archiveVersion = 1;
+	classes = {
+	};
+	objectVersion = 46;
+	objects = {
+
+/* Begin PBXBuildFile section */
+		7A04088C1DF9A2C7009C5BFA /* CHLObjcInstanceCommands.h in Headers */ = {isa = PBXBuildFile; fileRef = 7A04088A1DF9A2C7009C5BFA /* CHLObjcInstanceCommands.h */; };
+		7A04088D1DF9A2C7009C5BFA /* CHLObjcInstanceCommands.mm in Sources */ = {isa = PBXBuildFile; fileRef = 7A04088B1DF9A2C7009C5BFA /* CHLObjcInstanceCommands.mm */; };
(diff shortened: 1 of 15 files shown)


kastiglione:
For anyone wanting to try this at home, build the Xcode project, and get the path to the Chisel framework. On my machine it's at:

```
/Users/<me>/Library/Developer/Xcode/DerivedData/Chisel-<stuff>/Build/Products/Debug-iphonesimulator/Chisel.framework/Chisel
```

Then, in lldb do:

```
e @import Darwin
e dlopen("/path/to/Chisel", )
```

Then to call it, run:

```
script o=lldb.SBExpressionOptions(); o.SetTrapExceptions(False); o.SetTryAllThreads(False); o.SetTimeoutInMicroSeconds(10*1000000); lldb.frame.EvaluateExpression('(void)PrintInstances("<classname>", "<predicate>")', o).value
```

Where `<classname>` can be either a class name or a protocol name, and `<predicate>` is a string that can be parsed by `NSPredicate`.

This could be made easier using `regex command`:

```
command regex 's/findinstances (\w+)(?: +(.+))/o=lldb.SBExpressionOptions(); o.SetTrapExceptions(False); o.SetTryAllThreads(False); o.SetTimeoutInMicroSeconds(10*1000000); lldb.frame.EvaluateExpression('(void)PrintInstances("%1", "%2")', o).value/'
```

kastiglione:
I'll be doing follow up pull requests.

PR #199, opened 2017-08-26
Improve documentation

agisilaos:
This PR contains improvements to the current documentation. Before explaining the changes I would like to take a moment & explain why I made the PR.

So initially I had some problems getting started with chisel & installing it I thought I was the only one. But it turned out that other people as well had problems finding `.lldbinit` file, or even create it (A couple of issues created with such problems from users). 

These changes aim to bring clarity to the docs & help people that are new with such files/ installing projects like that get started in a few minutes. Into a bit more detail:

1) Added commands so the user can check if `.lldbinit` file exists & if it doesn't exist, to create it.

2) Added the command that needs to be added on the `.lldbinit` file when someone installs chisel with brew & I made such changes because sometimes user will not pay attention to the terminal after installing chisel with brew so now the docs will be improved & the experience of the user when he will install chisel much better.

These changes are one part of the overall adjustments/improvements that I'm thinking of doing on the documentation side of things. I would like to thank you for creating such an awesome tool for iOS engineers to work better while debugging.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,8 +15,8 @@ brew install chisel
 if `.lldbinit` file doesn't exist you can create it & open it by tapping on the terminal
 
  ```shell
- touch .lldbinit # creates file
- open .lldbinit # opens file
+ touch .lldbinit 
+ open .lldbinit 
 ```
 
 Then add the following line to your _~/.lldbinit_ file.
--- a/README.md
+++ b/README.md
@@ -11,10 +11,26 @@ For a comprehensive overview of LLDB, and how Chisel complements it, read Ari Gr
 brew update
 brew install chisel
 ```
+Then check if a  _~/.lldbinit_ file exists in your home directory by writing on the terminal the command below
 
-Then follow the instructions that Homebrew displays to add chisel to your _~/.lldbinit_.
+ ```shell
+ls ~/.lldbinit
+```
+if such file doesn't exist you can create it & open it by tapping on the terminal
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -11,10 +11,26 @@ For a comprehensive overview of LLDB, and how Chisel complements it, read Ari Gr
...
 
-Then follow the instructions that Homebrew displays to add chisel to your _~/.lldbinit_.
+ ```shell
+ls ~/.lldbinit
+```
+if such file doesn't exist you can create it & open it by tapping on the terminal
+
+ ```shell
+ touch .lldbinit //creates file
+ open .lldbinit //opens file
```
kastiglione: If someone doesn't know these commands, they might also not realize that the `//` comments are not supposed to be written. Generally I prefer not adding comments at the end of command lines, but if you feel strongly then they should at least be commented with `#` which is the shell syntax for comments.

agisilaos: Oups my bad there! Yeah makes sense.

agisilaos: I think the commands are self explanatory so I dropped the comments there. Good point

agisilaos:
I think I covered all of your points/ mentions @kastiglione. Please take a look & tell me what you think 😃

agisilaos:
Hey @kastiglione, any updates on this Pull Request? Is it ready to be merged? What do you think? Are there any other changes that you may want to suggest?

kastiglione:
Thanks for contributing this @agisilaos. Apologies for not being more of a prompt reviewer.

PR #200, opened 2017-09-04
Support debugging information overlay

hulizhen:
This command, `dio`, enables the feature that shows a float debugging information
overlay window, which could be very helpful for iOS develops and designers.

Continue program after executing this command, then tap the status bar with two
fingers at the same time to show the overlay window. You can dismiss it by tapping the upper-right `Dismiss` button after finishing.

The debugging information provided by this feature includes 'View Hierachy', 'VC Hierachy',
'Ivar Explorer', 'Measure', 'Spec Compare', 'System Color Audit', for now.

For more details, please refer to the original post: http://ryanipete.com/blog/ios/swift/objective-c/uidebugginginformationoverlay/

### DIFF
--- a/commands/FBDebugCommands.py
+++ b/commands/FBDebugCommands.py
@@ -12,6 +12,7 @@ def lldbcommands():
     FBFrameworkAddressBreakpointCommand(),
     FBMethodBreakpointCommand(),
     FBMemoryWarningCommand(),
+    FBDebuggingInformationOverlayCommand(),
   ]
 
 class FBWatchInstanceVariableCommand(fb.FBCommand):
@@ -184,3 +185,27 @@ class FBMemoryWarningCommand(fb.FBCommand):
 
   def run(self, arguments, options):
     fb.evaluateEffect('[[UIApplication sharedApplication] performSelector:@selector(_performMemoryWarning)]')
+
+class FBDebuggingInformationOverlayCommand(fb.FBCommand):
+  def name(self):
+    return 'dio'
+
+  def description(self):
+    return 'Enable the feature that shows a float debugging information overlay window. Continue program after executing this command, then tap the status bar with two fingers at the same time to show the overlay window. Only works on iOS device.'
+
+  def run(self, arguments, options):
+    if not objc.isIOSDevice():
+      print 'Sorry, but the ' + `self.name()` + ' command only works on iOS device.'
+      return
+
+    codeString = r'''
+    @import Foundation;
+    id DebugClass = NSClassFromString(@"UIDebuggingInformationOverlay");
+    
+    static dispatch_once_t onceToken;
+    dispatch_once(&onceToken, ^{
+    [DebugClass performSelector:NSSelectorFromString(@"prepareDebuggingOverlay")];
+    });
+    '''
+
+    lldb.debugger.HandleCommand("expression -lobjc -o -- " + codeString)
--- a/commands/FBDebugCommands.py
+++ b/commands/FBDebugCommands.py
@@ -188,24 +188,11 @@ class FBMemoryWarningCommand(fb.FBCommand):
 
 class FBDebuggingInformationOverlayCommand(fb.FBCommand):
   def name(self):
-    return 'dio'
+    return 'doverlay'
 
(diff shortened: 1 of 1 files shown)


commands/FBDebugCommands.py
```diff
@@ -184,3 +185,27 @@ def description(self):
...
+      print 'Sorry, but the ' + `self.name()` + ' command only works on iOS device.'
+      return
+
+    codeString = r'''
+    @import Foundation;
+    id DebugClass = NSClassFromString(@"UIDebuggingInformationOverlay");
+    
+    static dispatch_once_t onceToken;
+    dispatch_once(&onceToken, ^{
+    [DebugClass performSelector:NSSelectorFromString(@"prepareDebuggingOverlay")];
```
kastiglione: Can the overlay be dismissed? Or is it going to remain on screen for the rest of the time the app is running?

Why can't this be:

```objc
[UIDebuggingInformationOverlay prepareDebuggingOverlay]
```

hulizhen: 1. Yes, it can be dismissed by tapping the upper-right `Dismiss` button on the overlay window.
2. Because the `UIDebuggingInformationOverlay` class is a private class, we have to get the class object via Objective-C runtime functions.

kastiglione: 1. The use of `dispatch_once` implies the overlay can only be shown once. If it can be dismissed, what happens the second time `dio` is run?
2. While that's true when compiling code with clang, when evaluating code from lldb, private/internal classes and methods can be invoked with typical objc syntax.

kastiglione: After glancing at this [article on UIDebuggingInformationOverlay](http://ryanipete.com/blog/ios/swift/objective-c/uidebugginginformationoverlay/), it appears this code also needs a call to `[[UIDebuggingInformationOverlay overlay] toggleVisibility]`.

kastiglione: So I think this method could look something like this:

```python
fb.evaluateEffect("(void)[UIDebuggingInformationOverlay prepareDebuggingOverlay]")
fb.evaluateEffect("(void)[[UIDebuggingInformationOverlay overlay] toggleVisibility]")
```

commands/FBDebugCommands.py
```diff
@@ -184,3 +185,27 @@ def description(self):
 
   def run(self, arguments, options):
     fb.evaluateEffect('[[UIApplication sharedApplication] performSelector:@selector(_performMemoryWarning)]')
+
+class FBDebuggingInformationOverlayCommand(fb.FBCommand):
+  def name(self):
+    return 'dio'
```
kastiglione: Did you have any alternative names you considered? I think `dio` is good as an alias for a more memorable name, but I'm not sure what to call that. `debugoverlay`?

hulizhen: How about `doverlay`? Just like the command `dcomponents` which is already shipped with Chisel :)

kastiglione: `doverlay` 👍

commands/FBDebugCommands.py
hulizhen: Sorry for the late reply, I didn't see these ones :(

1. I just tried your one-line code with fb.evaluateEffect(), and it does works amazingly:

> fb.evaluateEffect("[UIDebuggingInformationOverlay prepareDebuggingOverlay]")

2. Actually, there are two ways to show the debugging overlay after calling `+[UIDebuggingInformationOverlay prepareDebuggingOverlay]`:

- Call `-[[UIDebuggingInformationOverlay overlay] toggleVisibility] `

-  Just tap the status bar with two fingers at the same time.

So the `toggleVisibility` method is not that necessary.

3. It seems that this does not work on iOS 11, but still works on older system. I will update the code when there is any workaround.

Thank you :)

kastiglione:
Apologies for letting this fall of my radar. @hulizhen if you're still ok with this, I'm happy to merge it.

hulizhen:
@kastiglione It's OK. But sadly, the command does not work on iOS 11 :(

PR #214, opened 2017-12-28
Add Makefile for installing framework

keith:
This allows you to run `make install` with optional environment variables
in order to build and install Chisel.framework.

### DIFF
--- a/Chisel/Makefile
+++ b/Chisel/Makefile
@@ -1,0 +1,10 @@
+PREFIX ?= /usr/local/lib
+
+install:
+	xcodebuild \
+		-scheme Chisel \
+		-configuration Release \
+		-sdk iphonesimulator \
+		install \
+		DSTROOT=/ \
+		INSTALL_PATH="$(PREFIX)"
--- a/Chisel/Chisel.xcodeproj/project.pbxproj
+++ b/Chisel/Chisel.xcodeproj/project.pbxproj
@@ -362,7 +362,7 @@
 				LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/Frameworks @loader_path/Frameworks";
 				PRODUCT_BUNDLE_IDENTIFIER = com.facebook.Chisel;
 				PRODUCT_NAME = "$(TARGET_NAME)";
-				SKIP_INSTALL = YES;
+				SKIP_INSTALL = NO;
 			};
 			name = Debug;
 		};
@@ -379,7 +379,7 @@
 				LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/Frameworks @loader_path/Frameworks";
 				PRODUCT_BUNDLE_IDENTIFIER = com.facebook.Chisel;
 				PRODUCT_NAME = "$(TARGET_NAME)";
-				SKIP_INSTALL = YES;
+				SKIP_INSTALL = NO;
 			};
 			name = Release;
 		};


Chisel/Makefile
```diff
@@ -0,0 +1,14 @@
...
+install:
+	xcodebuild \
+		-scheme Chisel \
+		-configuration Release \
+		-sdk iphonesimulator \
+		install \
+		DSTROOT="$(TEMPORARY_FOLDER)" \
+		INSTALL_PATH=/ \
+		SKIP_INSTALL=NO
+	install -d "$(PREFIX)"
```
kastiglione: Does this do anything beyond `mkdir -p`?

Chisel/Makefile
```diff
@@ -0,0 +1,14 @@
...
+TEMPORARY_FOLDER ?= /tmp/Chisel.dst
+
+install:
+	xcodebuild \
+		-scheme Chisel \
+		-configuration Release \
+		-sdk iphonesimulator \
+		install \
+		DSTROOT="$(TEMPORARY_FOLDER)" \
+		INSTALL_PATH=/ \
```
kastiglione: I don't know much about Xcode's install action. Why install into a temp path and then copy from there into the true destination?

kastiglione:
I should hopefully be able to accept this soon. In the mean time, I have a couple questions:

* Should we build a fat binary with x86 too?
* Should we code sign after copying into the destination?

I'm indifferent about the fat binary, I'm fine with being lazy and waiting for a request/need.

I don't know the ins and outs of code signing, but I hit some issues that frustrated me during development of this library. If I loaded the library from where it was originally built and code signed (within the build directory) and then later moved or copied the framework to some install location, then I would no longer be able to load the library from the installed location. The error messages were non-existent or cryptic at best, nothing that I could solve from google. Eventually I figured out that if I re-codesigned the installed library, then I wouldn't have those same loading problems. I guess what I'm asking is, is signing supposed to be post-install or can it be pre-install? Presumably pre-install, but the issues tell me that I don't really know how path comes into play with code signing verification.

Chisel/Makefile
keith: Removed with above changes

Chisel/Makefile
keith: Ah yes, nice I didn't re-think this after I made some other changes, updated

keith:
> Should we build a fat binary with x86 too?

This actually is building both right now, since it's a release build, `ONLY_ACTIVE_ARCH` is `NO`.

> Should we code sign after copying into the destination?

Good question. The `adhoc` signing that's happening when I run the install now is currently working for me, but I've also only tested on the simulator. If I remove that signature I do get this for `dlerror()`:

```
(char *) $3 = 0x00007fa68b400249 "dlopen(/usr/local/opt/chisel/lib/Chisel.framework/Chisel, 2): no suitable image found.  Did find:\n\t/usr/local/opt/chisel/lib/Chisel.framework/Chisel: required code signature missing for '/usr/local/opt/chisel/lib/Chisel.framework/Chisel'\n"
```

kastiglione:
I should be able to merge this next week.

Chisel/Makefile
```diff
@@ -0,0 +1,11 @@
+PREFIX ?= /usr/local/lib
+
+install:
+	xcodebuild \
+		-scheme Chisel \
+		-configuration Release \
+		-sdk iphonesimulator \
+		install \
+		DSTROOT="$(PREFIX)" \
+		INSTALL_PATH=/ \
```
kastiglione: According to the docs, the above two should be flipped. `DSTROOT=/` and `INSTALL_PATH=$(PREFIX)`.

Chisel/Makefile
```diff
@@ -0,0 +1,11 @@
...
+		SKIP_INSTALL=NO
```
kastiglione: Should this be set to `NO` in the `.xcodeproj`? I'm fine with it here either way, but I'm wondering why not have it default to this anyway.

keith: I'll set it in the project since there's no reason to archive it anyways

keith:
Addressed your feedback!

keith:
Thanks!

kastiglione:
You're welcome, thanks for the pull request. I'll create a pull request to add an implementation for the `findinstances` command some time later today.

PR #216, opened 2018-01-03
Implement findinstances command

kastiglione:
This pull requests adds the `findinstances` command, which completes the work from #197 and #214.

For details on `findinstances` users can run `help findinstances`. The brief summary is that `findinstances` can find all instances of a given class or protocol, and can filter those results, using a predicate expression.

If you had a class named `XXSocialUser`, then you could for example find a specific user by running `findinstances XXSocialUser username == 'curry'`.

### DIFF
--- a/commands/FBDebugCommands.py
+++ b/commands/FBDebugCommands.py
@@ -231,11 +231,14 @@ class FBFindInstancesCommand(fb.FBCommand):
     if not self.loadChiselIfNecessary():
       return
 
-    commands = arguments[0].strip().split(' ', 1)
-    query = commands[0]
-    if len(commands) > 1:
-      # TODO: Escape unescaped double quotes, and escape escapes
-      predicate = commands[1].strip()
+    # Unpack the arguments by hand. The input is entirely in arguments[0].
+    args = arguments[0].strip().split(' ', 1)
+
+    query = args[0]
+    if len(args) > 1:
+      predicate = args[1].strip()
+      # Escape double quotes and backslashes.
+      predicate = re.sub('([\\"])', r'\\\1', predicate)
     else:
       predicate = ''
     call = '(void)PrintInstances("{}", "{}")'.format(query, predicate)
--- a/commands/FBDebugCommands.py
+++ b/commands/FBDebugCommands.py
@@ -259,10 +259,18 @@ class FBFindInstancesCommand(fb.FBCommand):
       return True
 
     error = fb.evaluateExpressionValue('(char*)dlerror()')
+    errno = fb.evaluateExpressionValue('(int)errno')
     if error.unsigned != 0:
       print 'Error loading Chisel: ' + error.summary
+    elif errno.unsigned != 0:
+      error = fb.evaluateExpressionValue('(char*)strerror((int)errno)')
+      if error.unsigned != 0:
+        print 'Error loading Chisel: ' + error.summary
+      else:
+        print 'Error loading Chisel (errno {})'.format(errno)
     else:
       print 'Unknown error loading Chisel'
+
     return False
 
   def chiselLibraryPath(self):
--- a/commands/FBDebugCommands.py
+++ b/commands/FBDebugCommands.py
@@ -4,6 +4,8 @@ import lldb
 import fblldbbase as fb
 import fblldbobjcruntimehelpers as objc
 
+import sys
+import os
 import re
 
 def lldbcommands():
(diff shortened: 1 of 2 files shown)


kastiglione:
@keith wanna try this and let me know if it works for you?

keith:
I'm actually hitting the `Unknown error loading Chisel` code path, which is quite strange. I will try and debug more.

kastiglione:
Interesting. I can work with you on that.

kastiglione:
I wonder if `perror()‘ would shed any more light.

kastiglione:
Are you using `CHISEL_LIBRARY_PATH` or did you install it into the homebrew path? If using the environment variable, what path is the library?

kastiglione:
I added `errno` handling, which may help figure out the issue @keith.

keith:
```
error: error: Execution was interrupted, reason: EXC_BAD_ACCESS (code=50, address=0x123138000).
error: error: Execution was interrupted, reason: signal SIGKILL.
error: warning: got name from symbols: errno
error: cast from pointer to smaller type 'int' loses information
Unknown error loading Chisel
```

kastiglione:
Ah yes, `code=50`. This comes back to my question about code signing. That 50 is [`KERN_CODESIGN_ERROR`](https://twitter.com/kastiglione/status/808180520370638850).

So you installed via a local change to the chisel recipe?

If you re-`codesign` the library in it's installed location, does that fix the problem? The xcodebuild output will show the `codesign` command.

keith:
Ah yes. So the binary is signed from the homebrew install. I've added this command to make it work with `brew edit Chisel`:

```
    system "make", "-C", "Chisel", "install", "PREFIX=#{lib}"
```

But re-signing it does fix the problem, which is interesting.

keith:
Also I noticed if you run `findinstances` with no arguments, it fails:

```
(lldb) findinstances
  File "/usr/local/opt/chisel/libexec/fblldb.py", line 83, in runCommand
    command.run(args, options)
  File "/usr/local/Cellar/chisel/HEAD-76dfc8a/libexec/commands/FBDebugCommands.py", line 235, in run
    args = arguments[0].strip().split(' ', 1)
IndexError: list index out of range
```

It might be nice to have a nicer help message or something here

kastiglione:
Does homebrew build in a chroot and the copy out?

kastiglione:
> But re-signing it does fix the problem, which is interesting.

I wasted too much time over this. Where should the `codesign` command be added, the `homebrew` recipe?

kastiglione:
> run `findinstances` with no arguments, it fails

should be fixed in https://github.com/facebook/chisel/pull/216/commits/4ae2ba7afbbd84967883c7aa496121508f0d3dc7

kastiglione:
oh wait, just looked at the trace you gave, maybe not.

kastiglione:
New fix https://github.com/facebook/chisel/pull/216/commits/58659698604a9b7a2cd9eae6d14fac35597c501a. I actually ran it this time.

keith:
If we did add a codesign call in the formulae I'm not sure what signature you would use though

kastiglione:
Do you mean which identity? If not I don't follow.

keith:
Yes which signing identity for it to be valid, when I built manually the adhoc one that was used was valid.

kastiglione:
Shouldn't it always be ad hoc? Do simulator builds ever not use ad hoc? The `Makefile` you added will always use ad hoc, so resigning with ad hoc seems equal in intent.

kastiglione:
When you install using your local recipe (without further modifications), what is the result of:

`codesign -v -d /usr/local/Cellar/chisel/1.6.0/lib/Chisel.framework/Chisel`

kastiglione:
Note I'm not sure if my last question is useful for anything, feel free to ignore.

My thinking is that the homebrew recipe needs a post-install step that runs `codesign -s -`.

keith:
```
Executable=/usr/local/Cellar/chisel/HEAD-5865969/lib/Chisel.framework/Chisel
Identifier=com.facebook.Chisel
Format=bundle with Mach-O universal (i386 x86_64)
CodeDirectory v=20100 size=420 flags=0x2(adhoc) hashes=8+3 location=embedded
Signature=adhoc
Info.plist entries=20
TeamIdentifier=not set
Sealed Resources version=2 rules=13 files=2
Internal requirements count=0 size=12
```

kastiglione:
Looks the same as mine, except for the path. In my case, mine was installed using the `Makefile`, but I ran it myself, not homebrew. I'm wondering if homebrew runs it in a chroot then copies it to /usr/local, and if that breaks it. Try running the `make` command yourself, with `PREFIX=/usr/local/Cellar/chisel/HEAD-5865969/lib`.

kastiglione:
#217 is the fix for the code signing problems discussed here.

kastiglione:
works for @keith now

keith:
🎉

PR #217, opened 2018-01-04
Expose LD_DYLIB_INSTALL_NAME setting for `make`

kastiglione:
This change makes it so that `make` invocations can specify `LD_DYLIB_INSTALL_NAME`. At this time, this is done solely so that homebrew can set the value properly from the outset. Without this, homebew will modify the library after `make` has completed, but modifying a library after code signing means an error will occur when it's loaded at runtime.

### DIFF
--- a/Chisel/Makefile
+++ b/Chisel/Makefile
@@ -1,11 +1,16 @@
 PREFIX ?= /usr/local/lib
 
+export INSTALL_NAME = ""
+ifneq ($(LD_DYLIB_INSTALL_NAME), "")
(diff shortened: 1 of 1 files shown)


Chisel/Makefile
```diff
@@ -6,5 +6,6 @@ install:
 		-configuration Release \
 		-sdk iphonesimulator \
 		install \
+		LD_DYLIB_INSTALL_NAME="$(LD_DYLIB_INSTALL_NAME)" \
```
kastiglione: @keith Sir Make, how do I conditionally do this? If there's no `LD_DYLIB_INSTALL_NAME`, then I want to leave this out, so that it remains unset.

PR #219, opened 2018-01-05
Improve findinstances handling of Swift subclasses

kastiglione:
Calling `findinstances` currently requires the exact class name. For classes implemented in Swift, the name includes the module name, for example `SuperKit.DuperClass`. This change allows `findinstances` to be called with simply the class name (`DuperClass` in the previous example).

Note: This change does handle the case where more than one module contains a class by that name. For example, if two modules each define a class named `User`, then running `findinstances User` will list instances of both classes.

This pull request also includes a few other small improvements.

### DIFF
--- a/Chisel/Makefile
+++ b/Chisel/Makefile
@@ -1,7 +1,7 @@
 PREFIX ?= /usr/local/lib
 
-export INSTALL_NAME = ""
-ifneq ($(LD_DYLIB_INSTALL_NAME), "")
+export INSTALL_NAME =
+ifneq ($(LD_DYLIB_INSTALL_NAME),)
 	INSTALL_NAME = "LD_DYLIB_INSTALL_NAME=$(LD_DYLIB_INSTALL_NAME)"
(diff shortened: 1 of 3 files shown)


Chisel/Makefile
```diff
@@ -1,7 +1,7 @@
 PREFIX ?= /usr/local/lib
 
-export INSTALL_NAME = ""
-ifneq ($(LD_DYLIB_INSTALL_NAME), "")
+export INSTALL_NAME =
+ifneq ($(LD_DYLIB_INSTALL_NAME),)
```
kastiglione: @keith does this change look legit? The original here was still adding `LD_DYLIB_INSTALL_NAME=...` to `xcodebuild` command, but this change appears to fix that.

keith: yea seems fine, shouldn't have had the quotes in the first place I think, my bad.

keith:
Super nice improvement!

PR #228, opened 2018-01-31
Add missing use of fb.evaluateInputExpression()

### LINKED ISSUE #227: pjson not returning expected value with Swift
Using the latest version of chisel from Homebrew. When I use pjson on a `[String: String]` dictionary in Swift, it doesn't appear to work.

```
(lldb) pjson response 
error: error: use of undeclared identifier 'response'
error: error: use of undeclared identifier 'None'
None
```

Happy to help contribute a fix if this only works for Objective-C!

kastiglione:
This should improve a few more commands to be Swift compatible.

Fixes #227


dlo:
Thanks @kastiglione!

PR #230, opened 2018-02-23
Add enable&disable breakpoints by rg-expression.

sunbohong:


commands/FBDebugCommands.py
```diff
@@ -189,6 +191,58 @@ def run(self, arguments, options):
...
+def switchBreakpointState(expression,on):
+  ci = lldb.debugger.GetCommandInterpreter()
+  res = lldb.SBCommandReturnObject()
+
+  ci.HandleCommand('br list',res)
+  listStr = res.GetOutput()
+
+  br_list = listStr.splitlines()
+  expression_pattern = re.compile(r'{}'.format(expression),re.I)
+  id_pattern = re.compile(r'^\s*([1-9]\d*)(\.\d+)?',re.I)
```
kastiglione: What do you think about using the Python LLDB API, instead of parsing the output of `breakpoint list`?

For example, it would roughly look like:

```python
for breakpoint in target.breakpoint_iter():
  for location in breakpoint:
    if pattern.search(location.GetAddress().symbol.name):
      location.SetEnabled(False)
```

commands/FBDebugCommands.py
```diff
@@ -189,6 +191,58 @@ def run(self, arguments, options):
...
+            lldb.debugger.HandleCommand('br enable {}'.format(id))
+          else:
+            lldb.debugger.HandleCommand('br disable {}'.format(id))
+
+class FBMethodBreakpointEnableCommand(fb.FBCommand):
+  def name(self):
+    return 'benable'
+
+  def description(self):
+    return "Enable a set of breakpoints for a relative expression"
```
kastiglione: typo: should be "regular", not "relative"

commands/FBDebugCommands.py
```diff
@@ -189,6 +191,58 @@ def run(self, arguments, options):
...
+        if id:
+          print br
+          if on:
+            lldb.debugger.HandleCommand('br enable {}'.format(id))
+          else:
+            lldb.debugger.HandleCommand('br disable {}'.format(id))
+
+class FBMethodBreakpointEnableCommand(fb.FBCommand):
+  def name(self):
+    return 'benable'
```
kastiglione: lldb has an alias command named `rb` for **r**egex **b**reakpoint. I think it makes sense to call this `rbenable` and `rbdisable`, what do you think?

commands/FBDebugCommands.py
sunbohong:
```
  target = lldb.debugger.GetSelectedTarget()
  for breakpoint in target.breakpoint_iter():
    for location in breakpoint:
      if expression_pattern.search('{}'.format(location)):
        location.SetEnabled(on)
```
What do you think about using location? 
They have some different.
```
location	2.1: where = SunLigth`__66+[SUNNetService httpAsyncPostWithDictionary:completionBlock:]_block_invoke + 1984 at SUNNetService.m:37, address = 0x0000000104514dfc, unresolved, hit count = 0  Options: disabled 
address		SunLigth`__66+[SUNNetService httpAsyncPostWithDictionary:completionBlock:]_block_invoke + 1984 at SUNNetService.m:37
symbol		id = {0x00006db1}, range = [0x00000001001e063c-0x00000001001e0ec8), mangled="__66+[SUNNetService httpAsyncPostWithDictionary:completionBlock:]_block_invoke"
name 		__66+[SUNNetService httpAsyncPostWithDictionary:completionBlock:]_block_invoke

```
So that,we can use `benable disabled` to switch all breakpoints to `enable`,and also as follow
```

#use `benable disabled` to switch all breakpoints to `enable`
benable disabled

* benable ***address***
benable 0x0000000104514dfc

* benable ***filename***
benable SUNNetService.m 

* benable ***module***
benable UIKit
benable SunLigth 

```

kastiglione: Enabling/disabling via filename and module definitely seem useful!

commands/FBDebugCommands.py
```diff
@@ -189,6 +191,86 @@ def run(self, arguments, options):
     fb.evaluateEffect('[[UIApplication sharedApplication] performSelector:@selector(_performMemoryWarning)]')
 
 
+def switchBreakpointState(expression,on):
+  ci = lldb.debugger.GetCommandInterpreter()
+  res = lldb.SBCommandReturnObject()
```
kastiglione: These two lines are no longer used.

commands/FBDebugCommands.py
```diff
@@ -189,6 +191,86 @@ def run(self, arguments, options):
...
+
+  expression_pattern = re.compile(r'{}'.format(expression),re.I)
```
kastiglione: This can be `re.compile(expression, re.I)`

commands/FBDebugCommands.py
```diff
@@ -189,6 +191,86 @@ def run(self, arguments, options):
...
+
+  target = lldb.debugger.GetSelectedTarget()
+  for breakpoint in target.breakpoint_iter():
+    for location in breakpoint:
+      if expression_pattern.search('{}'.format(location)):
```
kastiglione: This also doesn't need the `format`, I think it can be just `.search(str(location))`.

commands/FBDebugCommands.py
```diff
@@ -189,6 +191,86 @@ def run(self, arguments, options):
...
+        print location
```
kastiglione: Nice addition 👍

commands/FBDebugCommands.py
```diff
@@ -189,6 +191,86 @@ def run(self, arguments, options):
...
+
+          * rbenable ***address***
+          benable 0x0000000104514dfc
+
+          #use `rbenable *filename*` to switch all breakpoints in this file to `enable`
+          benable SUNNetService.m 
+
+          #use `rbenable ***module(AppName)***` to switch all breakpoints in this module to `enable`
+          benable UIKit
+          benable Foundation
```
kastiglione: This documentation block uses `benable` where it should be `rbenable`.

commands/FBDebugCommands.py
```diff
@@ -189,6 +191,86 @@ def run(self, arguments, options):
...
+    return 'rbenable'
+
+  def description(self):
+    return """
+    Enable a set of breakpoints for a regular expression
+
+    Examples:
+
+          #use `rbenable disabled` to switch all breakpoints to `enable`
+          benable disabled
```
kastiglione: This doesn't appear to be implemented. I don't think it's necessary to implement, since `br dis` (`breakpoint disable`) does this already.

kastiglione:
@sunbohong I have just a few more small comments. Once those are resolved, I will merge this pull request. Thanks again.

commands/FBDebugCommands.py
sunbohong: since `br dis` (`breakpoint disable`)has already this.
Should we use `location.GetAddress()` to search?

sunbohong: image

```
location	57.1: where = libsystem_kernel.dylib`mach_msg_trap + 4, address = 0x00000001832ef564, resolved, hit count = 0 
address		libsystem_kernel.dylib`mach_msg_trap + 4
symbol		id = {0x0000024e}, range = [0x000000018095b560-0x000000018095b56c), name="mach_msg_trap"
name		mach_msg_trap
```

when i set breakpoint at this address(`0x0000024e`),only `symbol` contain `0x0000024e`.
So i want search `address` and `symbol`.How do you think?

kastiglione: I don't know what `id = {0x0000024e}` represents? The address of the breakpoint is in the `location`: `address = 0x00000001832ef564`.

sunbohong: i am sorry.i mean that i want search `0x1832ef564` ,not `0x00000001832ef564`

kastiglione: Do you mean you want `rbdisable 0x1832ef564` to disable the breakpoint whose address is `0x00000001832ef564`?

sunbohong: yes

kastiglione: Maybe have add special cased input handling. If the user's input expression matches `0x[[:xdigit:]]`, then when building the regex, add a `0*` after the `0x`. So `0x1832ef564` would become `0x0*1832ef564`.

kastiglione:
@sunbohong after thinking about it a little bit, I think `benable` and `bdisable` are good names for these commands. Disabling by address isn't really a regular expression, and I have an idea or two for extensions to this command, which are also not for regular expressions. One command I've built for myself in the past is to disable the current breakpoint when stopped, and I think `bdisable` is a good place to add that functionality.

commands/FBDebugCommands.py
```diff
@@ -189,6 +191,88 @@ def run(self, arguments, options):
...
+    return 'benable'
+
+  def description(self):
+    return """
+    Enable a set of breakpoints for a regular expression
+
+    Examples:
+
+          #use `benable disabled` to switch all breakpoints to `enable`
+          benable disabled
```
kastiglione: I don't think this should be documented. A builtin command already does this: `breakpoint enable` (or `br en` for short).

commands/FBDebugCommands.py
```diff
@@ -189,6 +191,88 @@ def run(self, arguments, options):
...
+    return 'bdisable'
+
+  def description(self):
+    return """
+    Disable a set of breakpoints for a regular expression
+
+    Examples:
+
+          #use `bdisable disabled` to switch all breakpoints to `disable`
+          bdisable disabled
```
kastiglione: Likewise, this is builtin.

sunbohong: ok,i will remove it

commands/FBDebugCommands.py
```diff
@@ -189,6 +191,83 @@ def run(self, arguments, options):
...
 
+def switchBreakpointState(expression,on):
+
+  expression_pattern = re.compile(r'{}'.format(expression),re.I)
+
+  target = lldb.debugger.GetSelectedTarget()
+  for breakpoint in target.breakpoint_iter():
+    if breakpoint.IsEnabled() != on and (expression_pattern.search(str(breakpoint))):
+      print str(breakpoint)
+      breakpoint.SetEnabled(on)
```
kastiglione: Did you find this was necessary in your experience? I'm curious what are the cases where these commands need to apply to the breakpoint itself instead of working only on the locations.

sunbohong: If we use `br disable 41` to off `41`,and `benable 41` only enable locations's state.
The `41` `41.1` will be ignore.

kastiglione:
Thanks @sunbohong!

PR #231, opened 2018-03-05
Make `paltrace` work on selected views

kastiglione:
Calling `_autolayoutTrace` prints the entire view tree, regardless of which view it's called on, but `_autolayoutTraceRecursively:` prints the tree of the view it's called on.


kastiglione:
works for me ¯\\\_(ツ)_/¯

Issue #222, opened 2018-01-25
visualize in Swift Error


ydms:
xcode 9, Swift4
in oc ok!

```
visualize self.view

error: error: use of undeclared identifier 'self'
Traceback (most recent call last):
  File "/usr/local/opt/chisel/libexec/fblldb.py", line 83, in runCommand
    command.run(args, options)
  File "/usr/local/Cellar/chisel/1.7.1/libexec/commands/FBVisualizationCommands.py", line 178, in run
    _visualize(arguments[0])
  File "/usr/local/Cellar/chisel/1.7.1/libexec/commands/FBVisualizationCommands.py", line 146, in _visualize
    if fb.evaluateBooleanExpression('(unsigned long)CFGetTypeID((CFTypeRef)' + target + ') == (unsigned long)CGImageGetTypeID()'):
TypeError: cannot concatenate 'str' and 'NoneType' objects
```

```
visualize `self.view`

error: error: use of undeclared identifier 'self'
```

```
visualize 0x7faf6b566d50

error: error: use of undeclared identifier 'CGFloat'
error: use of undeclared identifier 'CGRect'
error: missing '[' at start of message send expression
error: expected ']'
to match this '['
Traceback (most recent call last):
  File "/usr/local/opt/chisel/libexec/fblldb.py", line 83, in runCommand
    command.run(args, options)
  File "/usr/local/Cellar/chisel/1.7.1/libexec/commands/FBVisualizationCommands.py", line 178, in run
    _visualize(arguments[0])
  File "/usr/local/Cellar/chisel/1.7.1/libexec/commands/FBVisualizationCommands.py", line 152, in _visualize
    _showLayer('[(id)' + target + ' layer]')
  File "/usr/local/Cellar/chisel/1.7.1/libexec/commands/FBVisualizationCommands.py", line 103, in _showLayer
    width = float(fb.evaluateExpression('(CGFloat)(' + size + '.width)'))
TypeError: float() argument must be a string or a number
```

ollieatkinson:
Also seeing the same issue, tested on older version 1.5.0 with the same result:

```lldb
(lldb) visualize titleView
error: error: use of undeclared identifier 'CGFloat'
error: use of undeclared identifier 'CGRect'
error: missing '[' at start of message send expression
error: expected ']'
to match this '['
Traceback (most recent call last):
  File "/usr/local/opt/chisel/libexec/fblldb.py", line 83, in runCommand
    command.run(args, options)
[...]

Issue #208, opened 2017-11-09
border doesn't work with swift 3


aboelbisher:
Hi.
am trying to border a view (using swift 3 and Xcode 9.1) , but it keeps giving me this error : 
error: error: use of undeclared identifier 'CGFloat'
error: error: use of undeclared identifier 'CGColorRef'
error: expected ']'
to match this '['

kastiglione:
thanks, I'll take a look

xhruso00:
Reboot fix this issue for me.

xhruso00:
try in console 
e @import AppKit or
e @import UIKit

aboelbisher:
reboot what ? my macbook ? or just the Xcode ?

aboelbisher:
rebooting didn't fix the issue 
and importing gave me an error : 
error: <EXPR>:3:2: error: expected an attribute name
@import UIKit

and also #import UIKit gave the error :
error: error: use of undeclared identifier 'CGFloat'
error: error: use of undeclared identifier 'CGColorRef'
error: expected ']'
to match this '['

after calling "border self.view"

kastiglione:
This should be fixed by #232.

PR #232, opened 2018-03-06
Import UIKit to retry expression evaluation

### LINKED ISSUE #186: Unable To Add Border to UIView
Forgive me if I am not doing this right.

(lldb) pviews 0x7faee840af00
<_UILayoutGuide: 0x7faee840af00; frame = (0 736; 0 0); hidden = YES; layer = <CALayer: 0x608000023ae0>>
(lldb) border 0x7faee840af00
error: error: use of undeclared identifier 'CGFloat'
error: error: use of undeclared identifier 'CGColorRef'
error: expected ']'
to match this '['

PS. This is in a Swift project

kastiglione:
When an `EvaluateExpression()` call fails, inspect the resulting `SBError` to see whether the failure might be resolvable by importing missing modules. For now, this assumes importing UIKit will handle all cases. Chisel may need to import other modules too, but those can be added on a case by case basis.

This should handle many of the evaluation errors that happen in Swift projects.

Resolves #186 #208 #222

### DIFF
--- a/fblldbbase.py
+++ b/fblldbbase.py
@@ -40,6 +40,20 @@ class FBCommand:
   def run(self, arguments, option):
     pass
 
+def isSuccess(error):
+  # When evaluating a `void` expression, the returned value will indicate an
+  # error. This error is named: kNoResult. This error value does *not* mean
+  # there was a problem. This logic follows what the builtin `expression`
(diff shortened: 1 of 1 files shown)


kastiglione:
This fixes a problem with `alamborder`.

PR #239, opened 2018-04-04
Added FBPrintWindow command

PeteTheHeat:
I've been working with input accessory views quite a bit. They've been annoying to debug because they don't live in the main key window, they are hosted in the Keyboard's window. I added a short chisel command locally to make printing a window easy. Let me know if this is something useful for the library.

### DIFF
--- a/commands/FBPrintCommands.py
+++ b/commands/FBPrintCommands.py
@@ -52,6 +52,7 @@ class FBPrintViewHierarchyCommand(fb.FBCommand):
     return [
       fb.FBCommandArgument(short='-u', long='--up', arg='upwards', boolean=True, default=False, help='Print only the hierarchy directly above the view, up to its window.'),
       fb.FBCommandArgument(short='-d', long='--depth', arg='depth', type='int', default="0", help='Print only to a given depth. 0 indicates infinite depth.'),
+      fb.FBCommandArgument(short='-w', long='--window', arg='window', type='int', default="0", help='Specify the window to print a description of. Check which windows exist with "po (id)[[UIApplication sharedApplication] windows]".'),
     ]
 
   def args(self):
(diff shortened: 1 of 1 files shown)


kastiglione:
Thanks Pete. I like it. I have some thoughts.

This is the same as `pviews`, except `pviews` is hard coded to use `keyWindow`. What do you think about making this a flag on `pviews`? For example: `pviews -w 0`, `pviews -w 1`, etc.

kastiglione:
@PeteTheHeat what do you think about my suggestion?

PeteTheHeat:
Yeah that makes sense! Let me update it.

commands/FBPrintCommands.py
```diff
@@ -52,20 +52,27 @@ def options(self):
...
 
   def run(self, arguments, options):
     maxDepth = int(options.depth)
+    window = int(options.window)
     isMac = runtimeHelpers.isMacintoshArch()
 
-    if arguments[0] == '__keyWindow_dynamic__':
-      arguments[0] = '(id)[[UIApplication sharedApplication] keyWindow]'
-
+    if window > 0:
```
kastiglione: It looks like this check can be removed.

kastiglione: I see now why you have this. Never mind :)

commands/FBPrintCommands.py
```diff
@@ -20,6 +20,7 @@
 def lldbcommands():
   return [
     FBPrintViewHierarchyCommand(),
+    FBPrintWindow(),
```
kastiglione: Now that `pviews` supports `-w`, I don't think it's worth having a separate command. A user could get `pwindows` by adding this to their `~/.lldbinit`:

```
command alias pwindow pviews --window %1
```

commands/FBPrintCommands.py
PeteTheHeat: Yup I was hoping the if/else flow was clear, if a window is passed in, do some new stuff, else fall back to what the code was doing before.

commands/FBPrintCommands.py
PeteTheHeat: Yup, I tried to delete it but it came back in the merge somehow lol, I'm too used to hg these days. I'll remove it

kastiglione:
thanks!

PR #240, opened 2018-04-11
Fix issues found with sequence

kastiglione:
1. Call `HandleCommand()` without the context argument. This is because the context is essentially a cache, and its properties can be invalided by previous commands. For example if one command is `finish` then the `frame` property of the context will now be invalid and unusable. `HandleCommand()` without the context argument will use the current frame, for example.
2. Set the debugger to synchronous execution. Each command in the sequence must not run until the previous command has finished. If the debugger is in async mode, `HandleCommand()` can return asynchronously before the command has concluded.
3. Special case `continue` by calling directly `process.Continue()`. This is done as a workaround to `HandleCommand('continue')` not resulting in the expected result, it's as though the continue command is interrupted. Calling `Continue()` does result in the expected behavior.

### DIFF
--- a/commands/FBDebugCommands.py
+++ b/commands/FBDebugCommands.py
@@ -435,23 +435,53 @@ class FBSequenceCommand(fb.FBCommand):
     return 'Run commands in sequence, stopping on any error.'
 
   def lex(self, commandLine):
-    return commandLine.split(';')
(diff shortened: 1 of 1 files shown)


kolinkrewinkel:
Thanks for all the helpful code comments too!

PR #253, opened 2018-12-11
Update fblldb.py

prinsun:
fix Xcode 10 import script failed


kastiglione:
Thanks for the pull request. This looks fine to merge but I do wonder if there's a root problem elsewhere? I use Xcode 10 Chisel has worked for me. Also there have been no other reports of loading chisel in Xcode 10

PR #265, opened 2019-08-16
Remove little used, undocumented, and broken `pca`

### LINKED ISSUE #264: 'pca' command not work
I tried 'pca' command on backboard/Springboard/my app, all not work.

(lldb) pca
error: error: Execution was interrupted, reason: signal SIGABRT.

This not work too

(lldb) po [NSString stringWithCString:(char *)CARenderServerGetInfo(0, 2, 0)]
error: Execution was interrupted, reason: internal ObjC exception breakpoint(-3)..

kastiglione:
Closes #264

### DIFF
--- a/commands/FBPrintCommands.py
+++ b/commands/FBPrintCommands.py
@@ -104,17 +104,6 @@ class FBPrintViewHierarchyCommand(fb.FBCommand):
 
       print description
 
-class FBPrintCoreAnimationTree(fb.FBCommand):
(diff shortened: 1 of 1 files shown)


kolinkrewinkel:
In light of the fact that it's been broken – yeah, we should probably remove it. I don't have opposition to it being included in the project, but unless we have a fix in mind / time allocated to fix it...best not to have broken stuff lying around IMO.

PR #266, opened 2019-09-25
Make chisel compatible with Python 3

liuliu:
Xcode 11 shipped with Python 3 now. Unfortunately, chisel was written in
Python 2 style and will fail load for lldb. This PR fixed the Python 3
compatibility issues.


commands/FBDebugCommands.py
```diff
@@ -423,9 +423,9 @@ def isHeap(addr):
 
     allocations = (addr for addr in pointers if isHeap(addr))
     for addr in allocations:
-        print >>self.result, '0x{addr:x} {path}'.format(addr=addr, path=pointers[addr])
```
liuliu: I need some advice on this.

natansh: This might help.

https://stackoverflow.com/questions/34385674/what-does-print-do-in-python

I suspect this means we should write it as

`print('0x{addr:x} {path}'.format(addr=addr, path=pointers[addr], file=self.result)`

commands/FBDebugCommands.py
```diff
@@ -473,7 +473,7 @@ def run_command(self, interpreter, command):
     ret = lldb.SBCommandReturnObject()
     interpreter.HandleCommand(command, ret)
     if ret.GetOutput():
-      print >>self.result, ret.GetOutput().strip()
```
kolinkrewinkel: same approach as Natansh commented above

commands/FBDebugCommands.py
kastiglione: Yep

kastiglione:
Do you want chisel to remain compatible with python 2 going forward? Or should there by a tagged or branch that people can checkout if the want Python 2 support?

liuliu:
I believe the update itself is still compatible with Python 2 (I need to double check). Going forward, we will update our dev machines pretty aggressively to Xcode 11, thus, ongoing Python 2 support is not important to us. It is up to you guys to decide when you want to remove Python 2 support (considering macOS will stop shipping Python 2.7 in next major OS update and EOL is 2020).

kolinkrewinkel:
Yeah, I think based on the changes we made we're still (temporarily) Python 2 compatible. I'm okay with us becoming non-compatible though eventually, since both Apple and Facebook are moving pretty aggressively towards 3.

kastiglione:
> we're still (temporarily) Python 2 compatible

I think the use of `print(..., file=...)` needs changes to make it python 2 compatible. For python 2, it would have include `from __future__ import print_function`.

commands/FBDebugCommands.py
```diff
@@ -7,6 +7,8 @@
 import sys
 import os
 import re
+# Can be removed when Python 2 support is removed.
+from __future__ import print_function
```
kastiglione: I thought this would raise an exception under python 3, but it doesn't so that's good.

lsavino: I picked up this branch to run with Xcode 11 today (thank you @liuliu!) and got an error that this `__future__ import` needs to be at the top of the file. Moving it above the other imports silenced the warning.

liuliu:
LMK if there is anything else you need to update.

liuliu:
Doesn't seem like any conflicts, not sure if any other changes required after this merged in though given the latest tip. How to initiate merge from here?

kolinkrewinkel:
Oh I see, it's because there were conflict resolutions across multiple commits – if I squash, which is what I wanted to do anyways, it's all good!

Issue #292, opened 2021-04-24
bmessage  cause a error message


sunbohong:
```
(lldb) bmessage +[KKDic load]
Setting a breakpoint at +[NSObject load] with condition (void*)object_getClass((id)$x0) == 0x000000010071f878
error: Function name regular expression could not be compiled: repetition-operator operand invalid
```

kolinkrewinkel:
Hi @sunbohong , sorry for the (very large) delay in getting to this. Thanks for catching this, seems like a really useful fix! A few comments I think would be great to address before proceeding:
- Would be awesome to update the title of the commit to something more descriptive of the change itself for people looking through the commit log in the future.

PR #293, opened 2021-04-24
Update FBDebugCommands.py

sunbohong:
Fix https://github.com/facebook/chisel/issues/292

### DIFF
--- a/commands/FBDebugCommands.py
+++ b/commands/FBDebugCommands.py
@@ -168,7 +168,9 @@ class FBMethodBreakpointCommand(fb.FBCommand):
 
         methodIsClassMethod = methodTypeCharacter == "+"
 
-        if not methodIsClassMethod:
+        if methodIsClassMethod:
+            methodTypeCharacter = "\+"
+        else:
             # The default is instance method, and methodTypeCharacter
             # may not actually be '-'.
             methodTypeCharacter = "-"


commands/FBDebugCommands.py
```diff
@@ -168,7 +168,9 @@ def run(self, arguments, options):
 
         methodIsClassMethod = methodTypeCharacter == "+"
 
-        if not methodIsClassMethod:
+        if methodIsClassMethod:
+            methodTypeCharacter = "\+"
```
kolinkrewinkel, line 172: I think later on on L245, there's a place where we use the method type character in a non-regex context. Perhaps we can have a separate variable for the original methodTypeCharacter, as well as an escaped one we use when constructing regular expressions?

kolinkrewinkel (changes requested):
Hi @sunbohong , sorry for the (very large) delay in getting to this. Thanks for catching this, seems like a really useful fix! :)

A few comments I think would be great to address before proceeding:
- Would be awesome to update the title of the commit + PR to something more descriptive of the change itself for people looking through the commit log in the future.
- Attached a comment inline about a place where I think the escaping may cause an issue when used in a non-regex context.

commands/FBDebugCommands.py
kolinkrewinkel: As it turns out, I commented this but didn't share it with my colleague who approved the change. If you'd like to submit a follow-on, that'd be great. If not, I can take care of it eventually. :)

PR #303, opened 2022-03-10
Updates README.md to note brew path difference on M1 macs

rpassis:
As per title, `brew` is installed  on a different path on Apple Silicon/M1 machines as per docs [here](https://docs.brew.sh/FAQ#why-should-i-install-homebrew-in-the-default-location)

This PR updates the README to indicate that.


kolinkrewinkel (approved):
Thanks for adding this note!

